PR #3, opened 2017-03-23
Setup opam wrappers in created dockers

AltGr:
This should take the wrappers in `<opam_source>/shell/wrap-*.sh`, when they
exist (newer opams only) install them and set them up for the new opam.

This will enforce Linux namespaces upon building and installing the packages,
preventing them from doing network access when they shouldn't, writing files
where they shouldn't, etc.

NOTE 1: this is untested (but the namespace wrappers are)

NOTE 2: some current compiler definitions (e.g. 
https://github.com/ocaml/opam-repository/blob/master/compilers/4.03.0/4.03.0/4.03.0.comp) 
are broken, according to opam rules, and won't pass this. They need fixing,
i.e. not write to PREFIX before `make install`.

### DIFF
--- a/src/dockerfile_opam.ml
+++ b/src/dockerfile_opam.ml
@@ -69,8 +69,23 @@ let opam_init
 
 let install_opam_from_source ?prefix ?(branch="1.2") () =
   run "git clone -b %s git://github.com/ocaml/opam /tmp/opam" branch @@
-  Linux.run_sh "cd /tmp/opam && make cold && make%s install && rm -rf /tmp/opam"
+  let wrappers_dir = match prefix with
+  | None -> "/usr/local/share/opam"
+  | Some p -> Filename.concat p "share/opam"
+  in
+  Linux.run_sh
+    "cd /tmp/opam && \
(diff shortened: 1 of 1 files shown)


AltGr:
Ping ?

avsm:
Sorry about the delay -- I'm rebuilding all the containers with this change and will update here when they're active.

AltGr:
Be aware that note 2 above may break a lot of compilers, if not fixed

avsm:
I think that's ok for the containers -- we don't really test the compilers at the moment in CI, so this is a good excuse to break them and then fix them. I dont think it'll regress anything.

avsm:
Also it will only break the OPAM2 ones, which aren't something anyone is depending on that I know of.

avsm:
This is deployed now in ocaml/opam-dev, but the rc file is not enabled by default as it requires a privileged container to work by default. I've filed https://github.com/ocaml/opam/issues/2917 as the feature request to make it easier to activate these sandboxes in the Docker-based CI.

PR #7, opened 2017-09-14
Copy instead of build

samoht:
The ocaml/ci-opam image has to be manually pushed by using the script added in https://github.com/ocaml/ocaml-ci-scripts/pull/183

The good thing about this change remove all the build/packaging constraints that we had on ocaml-ci-scripts (e.g. simple build system, no dependency, etc) as this is now a static binary which is copied. The bad part is that it's a bit harder to update the binary with a custom `ci-opam` (e.g. `FORK_USER` and `FORK_BRANCH` will not work anymore). I think that's fine but feedback appreciated.

### DIFF
--- a/src-opam/dockerfile_distro.ml
+++ b/src-opam/dockerfile_distro.ml
@@ -20,7 +20,7 @@ open Dockerfile
 open Dockerfile_opam
 module Linux = Dockerfile_linux
 
-type t = [ 
+type t = [
   | `Alpine of [ `V3_3 | `V3_4 | `V3_5 | `V3_6 | `Latest ]
   | `CentOS of [ `V6 | `V7 ]
   | `Debian of [ `V10 | `V9 | `V8 | `V7 | `Stable | `Testing | `Unstable ]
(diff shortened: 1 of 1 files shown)


avsm:
This seems ok to me. I'll hook in a datakit ci that can push to all the images systematically..

src-opam/dockerfile_distro.ml
```diff
@@ -229,7 +229,7 @@ let apt_opam ?pin ?opam_version ?compiler_version labels distro tag =
     Linux.Git.init () @@
     opam_init ?compiler_version ~need_upgrade () @@
     (match pin with Some x -> run_as_opam "opam pin add %s" x | None -> empty) @@
-    run_as_opam "opam install -y depext travis-opam" @@
+    Dockerfile.copy ~from:"ocaml/ci-opam" ~src:["/ci-opam"] ~dst:"/usr/bin" () @@
```
avsm: does this really work?  I thought you had to specify a previous build stage, not an arbitrary other image...

samoht: See https://docs.docker.com/develop/develop-images/multistage-build/#use-an-external-image-as-a-stage :-)

PR #8, opened 2018-03-01
Fix and use Dockerfile_linux.*.dev_packages internally

kit-ty-kate:
This fixes the lack of ```Graphics``` module in the ```ocaml/opam2``` images.
It also uses an halfish-unused functions that could have fix the issue in the first place.

I'll be testing the result of those changes during the night on the ARM64 machine.

See log output of https://github.com/ocaml/opam-repository/pull/11504 and several others for the original issue with ```Graphics```.

### DIFF
--- a/src-opam/dockerfile_linux.ml
+++ b/src-opam/dockerfile_linux.ml
@@ -57,7 +57,8 @@ module RPM = struct
     run "chmod 700 .ssh"
 
   let dev_packages ?extra () =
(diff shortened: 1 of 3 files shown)


jpdeplaix:
Oh, maybe the CIs already in place here will do the job then.

jpdeplaix:
Mmh the Travis failure comes from the fact that ```ocam-dockerfile``` uses a dev version of ``%ocaml-version```

kit-ty-kate:
I also added ```patch``` on Alpine as it fixes at least this issue: https://github.com/ocaml/opam/issues/3227

avsm:
thanks!

PR #10, opened 2019-07-12
Mark Debian Stable as being Debian 10

kit-ty-kate:
The latest `debian-stable` docker image is still using Debian 9 instead of Debian 10 (now stable)


avsm:
good catch, thanks!

PR #13, opened 2019-08-05
Fix Travis

talex5:
Error was:

    Fatal error: exception Failure("No opam file found for my-package, aborting.")

Also, test on 4.08.


avsm:
Soon to be replaced with ocurrent :-)

PR #14, opened 2019-09-17
Add the Alpine testing repository

talex5:
This allows opam depexts to refer to testing packages with e.g.

    ["capnproto-dev@testing"] {os-distribution = "alpine"}


XVilka:
Totally agree. Recently met the problem that `lzop` tool comes only in the testing repository.

PR #16, opened 2019-11-26
Add Fedora 31 and Ubuntu 19.10

XVilka:
They were released recently, Fedora with 4.08.1 coming out of the box.


XVilka:
@avsm could you please take a look at this one?

avsm:
Perhaps you could provide some basic information with your PR to help with this, such as how you tested the new distros.  Otherwise I'll just be essentially starting from scratch when evaluating it.

talex5:
I just tested this by merging it into my branch and building the Docker images. The new Ubuntu one seems to work, but Fedora 31 failed because it was missing `m4` and `diffutils`. I also got some `find: command not found` warnings while building OCaml.

This commit fixed it for me https://github.com/talex5/ocaml-dockerfile/commit/cbb96c6f67ff884f459933f3ca9432caa854da90.

PR #18, opened 2019-12-25
Add missing AppVeyor config

XVilka:
To avoid it failing all the time. Also added 4.09 in Travis


avsm:
Will be fixed by an ocurrent Windows CI   (#32 )

PR #19, opened 2020-05-14
Install epel-release on CentOS

talex5:
This allows opam projects to depend on packages in the extra packages repository (e.g. `capnproto`).


kit-ty-kate:
Mh, I'm not sure this should be there. Opam already has a special case for installing `epel-release`: https://github.com/ocaml/opam-depext/blob/master/depext.ml#L216. Why is this needed?

talex5:
Ah thanks! I've made a PR to use that instead: https://github.com/ocaml/opam-repository/pull/16447

PR #20, opened 2020-05-31
Make the curl step fail if the download fails

talex5:
Otherwise, the download step succeeds but the tar step fails. Since the curl step succeeded, retrying just uses the cached version again.

    Step 6/56 : RUN curl -OL https://github.com/projectatomic/bubblewrap/releases/download/v0.3.1/bubblewrap-0.3.1.tar.xz
    Step 7/56 : RUN tar xf bubblewrap-0.3.1.tar.xz
    tar: This does not look like a tar archive


talex5:
Travis failed with:
```
[ERROR] The sources of the following couldn't be obtained, aborting:
          - ocaml-version.2.5.0:
```
I don't have permission to restart the build.

PR #23, opened 2020-10-10
Add support for archlinux

kit-ty-kate:
Docker now offers `archlinux` as an official image (https://hub.docker.com/_/archlinux) since last year. I remember having looked around for one not long before and I seem to have missed the news.

Anyway, I think that'd be really good to add it to the CI test array. https://github.com/ocaml-opam/opam-depext/pull/123 fixed the archlinux support and a new release of `opam-depext` is due soon. Anytime after that would be fine to merge this.

### DIFF
--- a/src-opam/dockerfile_distro.ml
+++ b/src-opam/dockerfile_distro.ml
@@ -20,6 +20,7 @@ open Astring
 
 type t = [
   | `Alpine of [ `V3_3 | `V3_4 | `V3_5 | `V3_6 | `V3_7 | `V3_8 | `V3_9 | `V3_10 | `V3_11 | `V3_12 | `Latest ]
+  | `Archlinux of [ `Latest ]
   | `CentOS of [ `V6 | `V7 | `V8 | `Latest ]
   | `Debian of [ `V10 | `V9 | `V8 | `V7 | `Stable | `Testing | `Unstable ]
   | `Fedora of [ `V21 | `V22 | `V23 | `V24 | `V25 | `V26 | `V27 | `V28 | `V29 | `V30 | `V31 | `V32 | `Latest ]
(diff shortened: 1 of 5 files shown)


kit-ty-kate:
All good and tested now.

Once this is merged there'll only be 4 linux distributions supported by `opam-depext` / `opam 2.1` that would still be missing/un-ci-tested:
- [oraclelinux](https://hub.docker.com/_/oraclelinux): would be easily added as we already have everything for yum-based distributions.
- [mageia](https://hub.docker.com/_/mageia): same thing
- [amazonlinux](https://hub.docker.com/_/amazonlinux) (support added in opam 2.1 I believe): I've never tried but I'm guessing it's the same thing as well.
- `Gentoo`: this one is a bit more tricky. There are "official" images (from gentoo's side) but it's not "official docker supported images" (as in, not in the default namespace): https://hub.docker.com/r/gentoo/stage3. There is a tiny bit more setup to do than other distributions (https://github.com/gentoo/gentoo-docker-images#using-the-portage-container-in-a-multi-stage-build) but all in all it wouldn't be too hard if we're willing to add it (I'd be up for it).

Once this is done we'll be able to test every single linux (base) distributions supported by `opam-depext` / `opam 2.1` using a simple docker image (currently under `ocurrent/opam`).
`opam-depext` / `opam 2.1` also supports:
- `macOS/Homebrew` (test using Travis-CI or Github Actions available)
- `macOS/MacPorts`
- `FreeBSD` (quick test using Travis-CI or the official livecd available)
- `OpenBSD` (quick test using [FuguIta](http://fuguita.org/)'s livecd available)
- `NetBSD`? (support removed in `opam 2.1`?)

cc @rjbou (this might be of interest to you, to know the state of the testable linux distributions with `opam-depext` / `opam 2.1`)

rjbou:
Thanks @kit-ty-kate !

kit-ty-kate:
`opam-depext` 1.1.4 is out. This is good to be merged. cc @avsm

avsm:
thanks!

PR #24, opened 2020-10-16
Distribution update

kit-ty-kate:
This PR:
* adds oraclelinux 8
* makes debian testing and unstable part of `latest_distros`
* update ubuntu latest to 20.04, fedora to 32, opensuse to 15.2, centos to 8
* removes tier2 support for alpine 3.11


avsm:
This PR doesn't compile, so I've taken portions of it in 19a7f3b7a88c9800d5a78b93861969d33b311732

- This also deprecates CentOS 7, which should remain a Tier 2 distro as its still used in some Xen places and isnt EOL
- Debian testing/unstable shouldn't be in latest_distros (they are for stable ones and Debian Stable is in there)

PR #25, opened 2020-10-16
New tier system

kit-ty-kate:
This PR sits on top of https://github.com/avsm/ocaml-dockerfile/pull/24

The current Tier system seems to me a bit arbitrary and doesn't have much basis on the actual state of the repository.
Furthermore I'm not sure where they are actually used (none of the CIs I know of use it)

Here I'm proposing to change it for deeper inclusion of `ocaml-dockerfile` in `ocaml-ci` and `ocaml-repo-ci`.

Tier 1 becomes: distributions well enough supported in opam-repository
Tier 2 becomes: distributions not well supported in opam-repository

`ocaml-ci` would use Tier 1 (which is incidently the list of distribution it is currently using + debian testing and debian sid)
`ocaml-repo-ci` would use Tier 1 and 2, which would have have as an effect to improve the state of Tier 2 platform for future inclusion in Tier 1.

### DIFF
--- a/src-opam/dockerfile_distro.ml
+++ b/src-opam/dockerfile_distro.ml
@@ -66,32 +66,38 @@ let distro_status (d:t) : status = match d with
   | `Alpine `V3_12 -> `Active `Tier1
   | `Alpine `Latest -> `Alias (`Alpine `V3_12)
   | `CentOS (`V6 | `V7) -> `Deprecated
-  | `CentOS `V8 -> `Active `Tier2
+  | `CentOS `V8 -> `Active `Tier1
   | `CentOS `Latest -> `Alias (`CentOS `V8)
   | `Debian (`V7 | `V8 | `V9) -> `Deprecated
   | `Debian `V10 -> `Active `Tier1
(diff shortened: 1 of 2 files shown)


avsm:
Thanks for the PR! 

> The current Tier system seems to me a bit arbitrary and doesn't have much basis on the actual state of the repository.

It's used to build the Docker base images; see https://github.com/ocurrent/docker-base-images/blob/master/src/conf.ml#L33

I can't quite tell easily what has changed in your PR from the old tiers, due to the whitespace reformatting. Any chance you could summarise them?

avsm:
I see what you want, and I'll combine it with the existing tier system.

- Tier 1 is currently used by docker-base-images and is "compile every possible supported variant of the OCaml compiler" and is only Debian Stable. It would be nice to add Alpine in here, but there are various known-broken combinations of musl and non-x86 upstream.
- Tier 2 can then be split up into distributions well enough supported in opam-repository
- Tier 3 for those not well supported in opam repository.

avsm:
See #30

PR #26, opened 2020-11-18
Enable the PowerTools repository on Centos 8 and above

kit-ty-kate:
Most of `*-devel` packages are on this new separate repository and opam-repo-ci fails very frequently on centos-8 because of that.


avsm:
Thanks, now live on base images

PR #31, opened 2020-12-16
Add CentOS SCL repo so we can get Cargo and friends on CI

avsm:
See https://www.softwarecollections.org/en/scls/rhscl/rust-toolset-7/
and https://github.com/ocaml/opam-repository/pull/17827


kit-ty-kate:
The rust-toolset-7 package sets does not seem add the binaries as part of the `$PATH` anyway. I doubt this is useful.
Though if that were, I think this would be better off in `opam-depexts` the same way `epel-release` is done than here.

PR #35, opened 2021-02-07
Install yum if not present

kit-ty-kate:
Should fix the current issue when building `oraclelinux-8`. The base image does not ship with yum anymore since https://github.com/oracle/container-images/commit/2f03b19e3230373aba31a91b315b1bbd1ff1998b


avsm:
thanks!

PR #40, opened 2021-03-13
Retrieve cygsympathy script from upstream master

dra27:

PR #41, opened 2021-03-15
Windows fixes

MisterDA:
New batch of fixes!
This in particular duplicates the base Windows image to have mingw and msvc images. The final design still may change.
Incorporates #40.

### DIFF
--- a/src-opam/dockerfile_opam.ml
+++ b/src-opam/dockerfile_opam.ml
@@ -214,17 +214,14 @@ let pacman_opam2 ?(labels=[]) ?arch distro () =
 
 
 (* Cygwin based Dockerfile *)
-let cygwin_opam2 ?(labels=[]) ?arch distro () =
-  Windows.Winget.build_form_source ?arch ~distro ()
-  @@ header ?arch distro @@ label (("distro_style", "cygwin") :: labels)
+let cygwin_opam2 ?(labels=[]) ~arch distro () =
+  Windows.Winget.build_from_source ~arch ~distro ()
+  @@ header ~arch distro @@ label (("distro_style", "cygwin") :: labels)
   @@ user "ContainerAdministrator"
   @@ Windows.install_vc_redist ()
-  @@ Windows.Winget.setup ()
-  @@ Windows.Winget.dev_packages ()
   @@ Windows.install_visual_studio_build_tools [
          "Microsoft.VisualStudio.Component.VC.Tools.x86.x64";
          "Microsoft.VisualStudio.Component.Windows10SDK.18362"]
-  @@ Windows.Winget.Git.init ()
   @@ begin
       let extra = Windows.Cygwin.msvc_packages () in
       let extra = Windows.Cygwin.mingw_packages ~extra () in
@@ -232,6 +229,9 @@ let cygwin_opam2 ?(labels=[]) ?arch distro () =
       let extra, t = Windows.Cygwin.ocaml_for_windows_packages ~extra () in
       Windows.Cygwin.setup ~extra () @@ t
     end
+  @@ Windows.Winget.setup ()
+  @@ Windows.Winget.dev_packages ()
+  @@ Windows.Cygwin.Git.init ()
   @@ Windows.cleanup ()
 
 let gen_opam2_distro ?(clone_opam_repo=true) ?arch ?labels d =
@@ -244,7 +244,10 @@ let gen_opam2_distro ?(clone_opam_repo=true) ?arch ?labels d =
      yum_opam2 ?labels ?arch ~yum_workaround ~enable_powertools d ()
   | `Zypper -> zypper_opam2 ?labels ?arch d ()
   | `Pacman -> pacman_opam2 ?labels ?arch d ()
-  | `Cygwin -> cygwin_opam2 ?labels ?arch d ()
+  | `Cygwin ->
+     match arch with
+     | None -> failwith "Windows requires an explicit architecture"
+     | Some arch -> cygwin_opam2 ?labels ~arch d ()
   in
   let clone = if clone_opam_repo then
     let url = Dockerfile_distro.(os_family_of_distro d |> opam_repository) in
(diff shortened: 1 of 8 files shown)


src-opam/dockerfile_windows.ml
```diff
@@ -75,16 +98,16 @@ module Cygwin = struct
     }
 
   let run_sh ?(cyg=default) fmt = ksprintf (run {|%s\bin\bash.exe --login -c "%s"|} cyg.root) fmt
+  let run_sh_ocaml_env ?(cyg=default) args fmt = ksprintf (run_sh "ocaml-env exec %s -- %s" args) fmt
```
dra27, line 101: The syntax for this is weird (I can never remember it...)
```suggestion
  let run_sh_ocaml_env ?cyg:(_ = default) args fmt = ksprintf (run_sh "ocaml-env exec %s -- %s" args) fmt
```
that's how you specify a default value for an optional parameter which you then don't use without triggering warning 27 😜

src-opam/dockerfile_opam.ml
```diff
@@ -244,7 +244,10 @@ let gen_opam2_distro ?(clone_opam_repo=true) ?arch ?labels d =
      yum_opam2 ?labels ?arch ~yum_workaround ~enable_powertools d ()
   | `Zypper -> zypper_opam2 ?labels ?arch d ()
   | `Pacman -> pacman_opam2 ?labels ?arch d ()
-  | `Cygwin -> cygwin_opam2 ?labels ?arch d ()
+  | `Cygwin ->
+     match arch with
+     | None -> failwith "Windows requires an explicit architecture"
```
dra27, line 249: Does it need an explicit arch? If so, this should be `invalid_arg`, not `failwith`

src-opam/dockerfile_windows.ml
```diff
@@ -20,6 +20,13 @@ open Printf
 
 let run_cmd fmt = ksprintf (run "cmd /S /C %s") fmt
 let run_powershell fmt = ksprintf (run {|powershell -Command "%s"|}) fmt
+let run_vc ~arch fmt =
+  let arch = match arch with
+    | `I386 -> "x86" | `X86_64 -> "amd64"
+    | `Aarch64 | `Aarch32 | `Ppc64le -> failwith "Cross-compiling isn't supported."
```
dra27, line 26: Cross-compiling is the wrong message here - it's an unsupported architecture. I have built OCaml on ARM64 Windows 10 using those tools - it's not cross-compilation!

src-opam/dockerfile_windows.ml
```diff
@@ -47,18 +54,34 @@ let install_visual_studio_build_tools ?(vs_version="16") ?(split=false) componen
...
 
-let append_path paths =
+let prepend_path paths =
   let paths = String.concat ";" paths in
   run {|for /f "tokens=1,2,*" %%a in ('reg query "HKLM\SYSTEM\CurrentControlSet\Control\Session Manager\Environment" /V Path ^| findstr /r "^[^H]"') do `
-       reg add "HKLM\SYSTEM\CurrentControlSet\Control\Session Manager\Environment" /V Path /t REG_EXPAND_SZ /f /d "%%c;%s"|} paths
+        reg add "HKLM\SYSTEM\CurrentControlSet\Control\Session Manager\Environment" /V Path /t REG_EXPAND_SZ /f /d "%s;%%c"|} paths
+
+type windows_port = [
+  | `Cygwin
```
dra27, line 63: `` `Cygwin`` is a Unix distribution, it shouldn't be here. I'm anticipating that we'll make it work by providing an equivalent OS base image (say to the Ubuntu ones) which just happens itself to be based on Windows.

src-opam/dockerfile_windows.ml
```diff
@@ -47,18 +54,34 @@ let install_visual_studio_build_tools ?(vs_version="16") ?(split=false) componen
...
-    | `X86_64 -> Some "mingw64c"
-    | _ -> None
-  else None
+    | `X86_64 | `I386 -> begin
+        match port with
+        | `Cygwin -> raise (Invalid_argument "Cygwin port is unsupported")
+        | `Mingw -> Some ("mingw" ^ bitness)
+        | `Msvc -> Some ("msvc" ^ bitness)
+      end
+    | _ -> raise (Invalid_argument "Unsupported architecture")
```
dra27, line 78:
```suggestion
    | _ -> invalid_arg "Unsupported architecture"
```

src-opam/dockerfile_windows.ml
```diff
@@ -120,23 +143,24 @@ module Cygwin = struct
   let msvc_packages ?(extra=[]) () = "make" :: "diffutils" :: extra
   let ocaml_for_windows_packages ?cyg ?(extra=[]) ?version:(version="0.0.0.2") () =
     let packages = "make" :: "diffutils" :: "mingw64-x86_64-gcc-g++" :: "vim" :: "git"
-                   :: "curl" :: "rsync" :: "unzip" :: "patch" :: "m4" :: extra in
+                   :: "curl" :: "rsync" :: "unzip" :: "patch" :: "m4" :: "gawk" :: extra in
```
dra27, line 146: gawk is a base package, so this shouldn't be necessary?

dra27 (commented):
#40's on its way in, so you can rebase without that commit with a few bits and bobs here there. Definitely skip Cygwin in this part for now - the aim is to spin out Cygwin into a container of its own which we won't maintain here (but which we may well then go on to _consume_ for the mingw/msvc images)

src-opam/dockerfile_opam.ml
MisterDA: iirc it's for the `ocaml-env` script. Although it claims to be able to detect 64 or 32 bits platforms, I've had cases where it defaulted to 32 bits and wouldn't load the correct compiler. So in this case, I'm requiring an explicit arch to be able to pass `--32` or `--64` later on.
I'll change to `invalid_arg`.

src-opam/dockerfile_windows.ml
MisterDA: Weird, I did not see the warning trigger. Noted, I didn't know that syntax. I even learned the shortcut `?(arg=value)` here :)

MisterDA: Oh, but it's a definitely a bug that the parameter is not used. It should be passed to the nested `run_sh` function.

dra27: Ah, so it is - I didn't stare at it hard enough to spot that!

PR #42, opened 2021-04-21
New super Windows fixes!

MisterDA:


dra27:
Can this be accompanied by an issue so we don't forget about it? I expect that what will happen is that by later in the summer when the Windows compiler are in opam-repository then we'll just be able to turn them back on here, rather than having to dig into the details in opam-repository-mingw but let's make sure we don't forget!

MisterDA:
This is good for merging, it will only require two minor updates when OCaml versions are released with the binutils 2.36 patch and when flexdll 0.39 enters Cygwin stable.

src-opam/dockerfile_windows.mli
```diff
@@ -76,8 +77,8 @@ module Cygwin : sig
   val update : ?cyg:cyg -> unit -> t
   (** Update Cygwin packages. *)
 
-  val cygwin_packages : ?cyg:cyg -> ?extra:string list ->
-                        ?flexdll_version:string -> unit -> string list * t
+  val cygwin_packages : ?extra:string list -> ?flexdll_version:string -> unit
+                        -> string list
   (** [cygwin_packages ?extra ()] will install the base development
      tools for the OCaml Cygwin port. Extra packages may also bep
```
avsm, line 83:
```suggestion
     tools for the OCaml Cygwin port. Extra packages may also be
```

PR #50, opened 2021-05-25
Better support for Windows 10 Updates

MisterDA:
This morning the Windows 21H1 update hit my laptop. It has not yet been published to Microsoft Docker repo, so we may want to wait before including this patch.

### DIFF
--- a/src-opam/dockerfile_distro.ml
+++ b/src-opam/dockerfile_distro.ml
@@ -18,6 +18,11 @@
 (** Distro selection for various OPAM combinations *)
 open Astring
 
+type win10_release = [
+  | `V1507 | `Ltsc2015 | `V1511 | `V1607 | `Ltsc2016 | `V1703 | `V1709
+  | `V1803 | `V1809 | `Ltsc2019 | `V1903 | `V1909 | `V2004 | `V20H2
+] [@@deriving sexp]
+
 type t = [
   | `Alpine of [ `V3_3 | `V3_4 | `V3_5 | `V3_6 | `V3_7 | `V3_8 | `V3_9 | `V3_10 | `V3_11 | `V3_12 | `V3_13 | `Latest ]
   | `Archlinux of [ `Latest ]
@@ -27,8 +32,8 @@ type t = [
   | `OracleLinux of [ `V7 | `V8 | `Latest ]
   | `OpenSUSE of [ `V42_1 | `V42_2 | `V42_3 | `V15_0 | `V15_1 | `V15_2 | `Latest ]
   | `Ubuntu of [ `V12_04 | `V14_04 | `V15_04 | `V15_10 | `V16_04 | `V16_10 | `V17_04 | `V17_10 | `V18_04 | `V18_10 | `V19_04 | `V19_10 | `V20_04 | `V20_10 | `V21_04 | `LTS | `Latest ]
-  | `Cygwin of [ `V20H2 ]
-  | `Windows of [`Mingw | `Msvc] * [ `V1809 | `V1903 | `V1909 | `V2004 | `V20H2 ]
+  | `Cygwin of win10_release
+  | `Windows of [`Mingw | `Msvc] * win10_release
 ] [@@deriving sexp]
 
 type os_family = [ `Cygwin | `Linux | `Windows ] [@@deriving sexp]
@@ -60,6 +65,7 @@ type status = [
   | `Deprecated
   | `Active of [ `Tier1 | `Tier2 | `Tier3 ]
   | `Alias of t
+  | `Not_available
 ] [@@deriving sexp]
 
 let distros = [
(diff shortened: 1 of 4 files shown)


dra27:
Ah, the 6-monthly march of progress (or May of progress)!

We definitely shouldn't _change_ the type of the `Cygwin constructor incompatibly. Equally, I think that once we're building a Cygwin image itself, we're going to want it multi-"arch" in the same way as the mingw/msvc images. So I think it's time to add `type win10_update = [ `V1809 | `V1903 | `V1909 | `V2002 | `V20H2 | V21H1 ]` and `win10_update_to_string` and `win10_update_of_string` and use that for the `` `Cygwin`` and `` `Windows`` constructors.

The type may as well include all the previous releases so `` `V1507`` onwards (technically Windows 10 1507 was not an update, but we can live with that 😉). The string conversions could possibly codenames as well - i.e. `"1507"` or `"Threshold 1"`, etc. The codename and version names are now in sync since 20H2.

src-opam/dockerfile_distro.ml
```diff
@@ -74,9 +74,9 @@ let distros = [
   `Ubuntu `V12_04; `Ubuntu `V14_04; `Ubuntu `V15_04; `Ubuntu `V15_10;
   `Ubuntu `V16_04; `Ubuntu `V16_10; `Ubuntu `V17_04; `Ubuntu `V17_10; `Ubuntu `V18_04; `Ubuntu `V18_10; `Ubuntu `V19_04; `Ubuntu `V19_10; `Ubuntu `V20_04; `Ubuntu `V20_10; `Ubuntu `V21_04;
   `Ubuntu `Latest; `Ubuntu `LTS;
-  `Cygwin `V20H2;
-  `Windows (`Mingw, `V1809); `Windows (`Mingw, `V1903); `Windows (`Mingw, `V1909); `Windows (`Mingw, `V2004); `Windows (`Mingw, `V20H2);
-  `Windows (`Msvc, `V1809); `Windows (`Msvc, `V1903); `Windows (`Msvc, `V1909); `Windows (`Msvc, `V2004); `Windows (`Msvc, `V20H2);
+  `Cygwin `V21H1;
+  `Windows (`Mingw, `V1809); `Windows (`Mingw, `V1903); `Windows (`Mingw, `V1909); `Windows (`Mingw, `V2004); `Windows (`Mingw, `V20H2); `Windows (`Mingw, `V21H1);
+  `Windows (`Msvc, `V1809); `Windows (`Msvc, `V1903); `Windows (`Msvc, `V1909); `Windows (`Msvc, `V2004); `Windows (`Msvc, `V20H2); `Windows (`Msvc, `V21H1);
```
dra27, line 79: We're going to make a mistake with this at some point:

```ocaml
let distros = ...
let windows_docker_bases = [`V1809; `V1903; `V1909; `V2004; `V20H2; `V21H1] in
let distros = distros @
  List.map (fun u -> `Cygwin u) windows_docker_bases @
  List.map (fun u -> `Windows (`Mingw, u)) windows_docker_bases @
  List.map (fun u -> `Windows (`Msvc, u)) windows_docker_bases
```

MisterDA:
I was thinking of adding that type.
I'd say we don't list updates that are not available as Docker images (the older ones). Are they going to be available in the future?

src-opam/dockerfile_distro.ml
```diff
@@ -109,14 +109,14 @@ let distro_status (d:t) : status = match d with
   | `Ubuntu ( `V12_04 | `V14_04 | `V15_04 | `V15_10 | `V16_04 | `V16_10 | `V17_04 | `V17_10 | `V18_10 | `V19_04 | `V19_10 ) -> `Deprecated
   | `Ubuntu `LTS -> `Alias (`Ubuntu `V20_04)
   | `Ubuntu `Latest -> `Alias (`Ubuntu `V20_10)
-  | `Cygwin `V20H2 -> `Active `Tier3
+  | `Cygwin `V21H1 -> `Active `Tier3
   | `Windows (_, _) -> `Active `Tier3
```
dra27, line 113: We need a `win10_update_supported` function for this. 1903, for example, ended support in November. I presume that the 1809 Docker images are receiving LTSC updates, but that's worth checking (by looking at the build number and checking the Latest Cumulative Update info from Microsoft). It may anyway be worth a having that function return a level of support - i.e. ``[`Consumer | `Enterprise | `LTSC]`` (see the [helpful table on Wikipedia](https://en.wikipedia.org/wiki/Windows_10_version_history#Channels)).

src-opam/dockerfile_distro.ml
```diff
@@ -109,14 +109,14 @@ let distro_status (d:t) : status = match d with
...
 
 let latest_distros =
   [ `Alpine `Latest; `Archlinux `Latest; `CentOS `Latest;
     `Debian `Stable; `OracleLinux `Latest; `OpenSUSE `Latest;
     `Fedora `Latest; `Ubuntu `Latest; `Ubuntu `LTS;
-    `Cygwin `V20H2; `Windows (`Mingw, `V20H2); `Windows (`Msvc, `V20H2)]
+    `Cygwin `V21H1; `Windows (`Mingw, `V21H1); `Windows (`Msvc, `V21H1)]
```
dra27, line 119: This is the tag we use for the unmarked version? I think policy-wise it might be worth lagging one month behind release - i.e. allow the first LCU update. Equally, if there's a problem we can rebuild the docker image, so I'm not certain about that!

src-opam/dockerfile_distro.ml
```diff
@@ -217,7 +217,7 @@ let builtin_ocaml_of_distro (d:t) : string option =
   |`OpenSUSE `V15_2 -> Some "4.05.0"
   |`OracleLinux `V7 -> Some "4.01.0"
   |`OracleLinux `V8 -> Some "4.07.0"
-  |`Cygwin `V20H2 -> Some "4.10.0"
+  |`Cygwin `V21H1 -> Some "4.10.0"
```
dra27, line 220: I'd change this to `None` on the basis that the OCaml in Cygwin64 at present is broken!

src-opam/dockerfile_distro.ml
```diff
@@ -291,17 +291,19 @@ let tag_of_distro (d:t) = match d with
...
   |`Windows (`Mingw, `V1909) -> "windows-mingw-1909"
   |`Windows (`Mingw, `V2004) -> "windows-mingw-2004"
   |`Windows (`Mingw, `V20H2) -> "windows-mingw-20H2"
+  |`Windows (`Mingw, `V21H1) -> "windows-mingw-21H1"
   |`Windows (`Msvc, `V1809) -> "windows-msvc-1809"
   |`Windows (`Msvc, `V1903) -> "windows-msvc-1903"
   |`Windows (`Msvc, `V1909) -> "windows-msvc-1909"
   |`Windows (`Msvc, `V2004) -> "windows-msvc-2004"
   |`Windows (`Msvc, `V20H2) -> "windows-msvc-20H2"
+  |`Windows (`Msvc, `V21H1) -> "windows-msvc-21H1"
```
dra27, line 306: A similar refactoring with a function to convert the version I think is worth doing for safety, given that this will be changed every six months (and Microsoft change the naming frequently enough that I don't think we should generate the names!)

dra27 (commented):
Various comments - I think the code should be refactored a bit to allow for the need to update this every six months.

src-opam/dockerfile_distro.ml
MisterDA: I did not consider the level of support, but I think I got a sensible fix around this.

src-opam/dockerfile_distro.ml
```diff
@@ -18,6 +18,12 @@
 (** Distro selection for various OPAM combinations *)
 open Astring
 
+type win10_update =
+  [ `V1507 | `V1511 | `V1607 | `V1703 | `V1709 | `V1809 | `V1903 | `V1909 | `V2004 | `V20H2 | `V21H1 ] [@@deriving sexp]
+let win10_updates =
```
dra27, line 23: Could pedantically call this `win10_releases` to cover the fact that 1507 is the original?

MisterDA:
I think this is done, what do you think?

src-opam/dockerfile_distro.ml
```diff
@@ -74,10 +80,20 @@ let distros = [
...
-  `Cygwin `V20H2;
-  `Windows (`Mingw, `V1809); `Windows (`Mingw, `V1903); `Windows (`Mingw, `V1909); `Windows (`Mingw, `V2004); `Windows (`Mingw, `V20H2);
-  `Windows (`Msvc, `V1809); `Windows (`Msvc, `V1903); `Windows (`Msvc, `V1909); `Windows (`Msvc, `V2004); `Windows (`Msvc, `V20H2);
 ]
+let distros = List.fold_left (fun distros version ->
+                  `Cygwin version :: `Windows (`Mingw, version) :: `Windows (`Msvc, version) :: distros)
+                distros win10_releases
+
+(* https://docs.microsoft.com/en-us/virtualization/windowscontainers/deploy-containers/base-image-lifecycle *)
+let win10_release_status v ~servercore : status = match v with
```
dra27, line 89: Server Core is an implementation detail - I think what's wanted here is the support channel. You could have a variant `type channel = SemiAnnual | LongTerm` or we could just use a boolean `~ltsc`.

src-opam/dockerfile_distro.ml
```diff
@@ -74,10 +80,20 @@ let distros = [
...
+  | `V1507 | `V1511 -> `Deprecated
+  | `V1607 -> if servercore then `Active `Tier3 else `Deprecated
+  | `V1703 | `V1709 -> `Deprecated
+  | `V1809 -> `Active `Tier3
+  | `V1903 -> `Deprecated
+  | `V1909 | `V2004 | `V20H2 -> `Active `Tier3
+  | `V21H1 -> `Deprecated       (* Not yet available. *)
```
dra27, line 96: I think better for this to be `` `Active `Tier3`` and handle the fact the container isn't yet available elsewhere (either by delaying the merge or by having a special case in docker-base-images, or special-casing it below in distro_status).

dra27 (commented):
A little bike-sheddy, but if we settle the layout and API now it's just 6-monthly metadata updates!

src-opam/dockerfile_distro.ml
MisterDA: Changed that to explicitly list LTSC versions too. They're still supported, their semi-annual equivalent isn't and LTSC versions don't alias to semi-annual.

src-opam/dockerfile_distro.ml
MisterDA: Now we track the Microsoft support status of each version, and their availability in the Docker repository, so we correctly report that 21H1 has been released by Microsoft, but isn't available as a Docker image (yet).

avsm (approved):

PR #52, opened 2021-06-14
Update the opam 2.1 support to use the new 2.1 branch instead of master

kit-ty-kate:
opam 2.1 just branched from master so this PR is here to update dockerfile-opam to use this new branch instead of master

### DIFF
--- a/src-opam/dockerfile_opam.ml
+++ b/src-opam/dockerfile_opam.ml
@@ -110,20 +110,20 @@ let header ?arch ?maintainer ?img ?tag d =
   @@ shell
 
 (* Apk based Dockerfile *)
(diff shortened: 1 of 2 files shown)


dra27:
Given that there's going to be `opam-2.0`, `opam-2.1` and `opam-2.2` as soon as there's something meaningful on opam's master branch, would it be worth generalising `hash_opam_2_0:string -> hash_opam_2_1:string` to `opams:(string * string) list` so passing in a list of branchname, sha pairs? Otherwise the API changes again in a week or so?

kit-ty-kate:
yes indeed it is worth generalizing but I think this is ok if it will change again next week or so. I feel like this is better to split the urgent PR from the more prone to discussion API change.

avsm:
Agreed, we can make that change to also install opam-2.2 once this is rebuilt.

PR #54, opened 2021-06-29
Winget fixes

MisterDA:
Microsoft changed the release format of Winget and discontinued support for versions of Windows older than Windows 10 1809, so I disabled Winget entirely on these systems (previous versions of Winget are buggy anyway...).


avsm:
Lgtm!

PR #56, opened 2021-07-14
Add support for Windows 10 Latest Cumulative Update tags

dra27:
Windows Containers cannot start an image for a newer version of Windows and shouldn't attempt to start a container with a newer LCU (although the container should physically start).

On the second Wednesday of the month, `mcr.microsoft.com/windows:20H2`, etc. will all point to the LCU released the previous evening. The base-images should not immediately start pulling that until the workers have been updated - I think it's better therefore to refer to `mcr.microsoft.com/windows:20H2-KB5003637` (which is the 8 June 2021 LCU) which will always pull the same image. Once the workers are updated, we update the base-images builder with the new LCU tag.

This adds support for appending the KB number for the LCU for a given month. Microsoft only release Docker images for the official "update Tuesday" LCU, and not any out-of-band LCUs, so the OOB LCUs are not included. Since the KB number can differ by version, the LCU is exposed as a date and is then mapped when given a Windows 10 release to a KB number to append to the tag.

I've "structured" this so that the LCU gets added as the tag is being generated - i.e. it's not part of the Windows 10 version. While the different releases of Windows 10 are distinct (and we'll always want to support multiple releases at once) I cannot see the sense in supporting multiple LCUs.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -13,6 +13,7 @@ dev
 - Introduce a Cygwin image, and move Windows Mingw and Windows Msvc at
   the distro level. (@MisterDA #42)
 - Add edge and edgecommunity repositories to the Alpine image (@EwanMellor #48)
+- Add support for Windows 10 Latest Cumulative Update docker image tags (@dra27 #56)
 
 v7.1.0 2021-02-25 Cambridge
 ---------------------------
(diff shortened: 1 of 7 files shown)


talex5:
CI says: `Error (warning 8 [partial-match]): this pattern-matching is not exhaustive.`
(note: vendored libraries build with non-fatal warnings)

dra27:
Oops!

MisterDA:
LGTM!

PR #62, opened 2021-09-02
Ensure stray double-quotes don't get into PATH

dra27:
If the registry value for Path ends with a backslash than `\"` ends up in a command line which is interpreted as a double-quote character by the somewhat messed up rules of command line parsing in Windows! The "simplest" fix is to ensure this never happens. The actual fix would be to add an extra backslash to the end of command lines when there's an odd number of terminating backslashes - but try doing that before you've got `sed`!

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,13 @@
+unreleased
+----------
+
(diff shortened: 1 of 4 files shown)


MisterDA:
LGTM but I'm afraid a Great Old One might show up if I read the command out loud.

PR #63, opened 2021-09-07
Introduce Windows 10 LTSC 2022 and Windows Server image

MisterDA:
LTSC is a bit weird: it's not exactly an alias to 21H1, and 21H1
hasn't been released yet. For now, let's consider that they are
independent releases and that 21H1 isn't yet available.

Also introduce the "windows/server" image that replaces the "windows"
image starting with ltsc2022.

### DIFF
--- a/src-opam/dockerfile_distro.ml
+++ b/src-opam/dockerfile_distro.ml
@@ -34,7 +34,7 @@ let win10_current_lcu = `LCU20210914
 
 type win10_revision = win10_release * win10_lcu option [@@deriving sexp]
 
-let win10_lcus = [
+let win10_lcus : ('a * int * win10_release list) list = [
   `LCU20210914, 5005575, [`Ltsc2022];
   `LCU20210914, 5005565, [`V2004; `V20H2; `V21H1];
   `LCU20210914, 5005566, [`V1909];
@@ -59,7 +59,7 @@ let win10_lcus = [
   `LCU20210608, 5003687, [`V1507; `Ltsc2015];
 ]
 
-let win10_lcu_to_kb =
+let win10_lcu_to_kb : ((win10_lcu * win10_release), int option) Hashtbl.t =
   let t = Hashtbl.create 63 in
   let f (lcu, kb, vs) =
     let g v =
@@ -79,7 +79,7 @@ let win10_lcu_kb_number v lcu =
   try Hashtbl.find win10_lcu_to_kb (lcu, v)
   with Not_found -> None
 
-let win10_kb_number_to_lcu v kb =
+let win10_kb_number_to_lcu (v:win10_release) kb =
   match Hashtbl.find win10_kb_to_lcu (kb, v) with
   | lcu -> Some (v, lcu)
   | exception Not_found -> None
@@ -385,7 +385,7 @@ let rec win10_revision_to_string = function
 | (v, Some lcu) ->
     match win10_lcu_kb_number v lcu with
     | Some kb -> Printf.sprintf "%s-KB%d" (win10_release_to_string v) kb
-    | None -> invalid_arg "No KB for this Win10 revision"
+    | None -> Fmt.invalid_arg "No KB for this Win10 %s revision" (win10_release_to_string v)
 
 let win10_revision_of_string v =
   let v, lcu =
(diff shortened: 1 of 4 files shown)


dra27:
I don't think we should introduce the `` `Server`` tag - what Microsoft's really doing is _renaming_ the repo from `windows` to `windows/server` (by only adding new tags to `windows/server`). The point in their docs is that the `windows` repo has _always_ been Windows Server (+ a few extra binaries) not the Windows 10 Client. I think `win10_docker_base_image` encompasses the distribution of Windows Server which you're expecting: Nano Server, Core or "Full" and they've just changed the repo for Full in ltsc2022 from `windows` to `windows/server`. Does that make sense? At the moment, there are match cases missing to exclude say ``(`Server, `V2004)`` in `win10_docker_status` but I think the problem is the type, not the missing match cases!

MisterDA:
I agree with everything, in the end it's just a matter of renaming.

> At the moment, there are match cases missing to exclude say (`Server, `V2004) in win10_docker_status 

The compiler would have caught that and I mostly compile my code before sending it :smile: 
> `` | `Server, _ -> `Not_available``

dra27:
> The compiler would have caught that and I mostly compile my code before sending it 😄
> 
> > ``  | `Server, _ -> `Not_available ``

Oops - my bad. I completely missed that line! 🙂 The compiler wouldn't have caught it, though, because of the wildcards below.

src-opam/dockerfile_distro.ml
```diff
@@ -157,15 +159,16 @@ let win10_release_status v : win10_release_status = match v with
   | `V1909 -> `Deprecated
   | `V2004
   | `V20H2
-  | `V21H1 -> `Active
+  | `V21H1 | `Ltsc2022 -> `Active
 
-let win10_latest_release = `V21H1
+let win10_latest_release = `Ltsc2022
```
dra27, line 164: This is the latest release of Windows 11! It's all a bit odd, given the "last" release of Windows 10 clearly isn't, but I think this can stay at `` `V21H1`` (especially given that we don't use it IIRC?)

src-opam/dockerfile_distro.ml
```diff
@@ -179,7 +182,7 @@ let win10_docker_status (base : win10_docker_base_image) v : status =
   | `NanoServer, `V1607 -> `Deprecated
   | _ -> `Not_available
 
-let win10_latest_image = `V20H2
+let win10_latest_image = `Ltsc2022
```
dra27, line 185: Conversely this one is fine, if now slightly strangely named!

src-opam/dockerfile_distro.ml
MisterDA: Changed back to `` `V21H1 ``.

src-opam/dockerfile_distro.ml
MisterDA: Leaving it as-is, then!

dra27:
LGTM - let's test this as part of the docker-base-images update and then merge.

dra27:
The repo needs updating here:

https://github.com/avsm/ocaml-dockerfile/blob/313b1c9b91b49bdc5b536f2992f079ae91047611/src-opam/dockerfile_windows.ml#L162-L166

MisterDA:
Thanks, fixed by exposing a function and a type to get the image and tag for Windows base container images.

src-opam/dockerfile_distro.mli
```diff
@@ -152,6 +153,22 @@ val latest_tag_of_distro : t -> string
   regularly rewritten to point to any new releases of the
   distribution. *)
 
+type win10_docker_base_image = [
+  | `NanoServer                 (** An ultralight Windows image. *)
+  | `ServerCore                 (** Medium in size. *)
+  | `Windows                    (** Largest and full Windows API support. *)
```
dra27, line 159:
```suggestion
  | `NanoServer (** Windows Nano Server *)
  | `ServerCore (** Windows Server Core *)
  | `Windows    (** Windows Server "with Desktop Experience" *)
```

dra27 (commented):
Docs tweak only - look great!

dra27:
ltsc2022/Windows 11 images now build 🥳

PR #67, opened 2021-09-28
distro updates for alpine 3.14/ubuntu 21.10 and refresh

avsm:
- Add Alpine 3.14 and Ubuntu 21.10
- Move Fedora 33 and OpenSUSE 15.2 to deprecated and Alpine 3.13 to Tier 2.
- Latest Fedora is now Fedora 34

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,8 +1,12 @@
 unreleased
 ----------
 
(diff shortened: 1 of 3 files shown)


dra27:
OK to delete the branch?

PR #73, opened 2021-11-02
Switch all uses of the git:// protocol to https:// (Github disabled git://)

kit-ty-kate:
https://github.blog/2021-09-01-improving-git-protocol-security-github/


talex5 (approved):

dra27:
From the [blog post](https://github.blog/2021-09-01-improving-git-protocol-security-github/#when-are-these-changes-effective), it's not entirely clear whether they _meant_ to turn off the git protocol on 2nd as part of that brownout, but they are supposed to be doing that in January, so at least we're ready!

PR #76, opened 2021-11-16
Add opam master (opam-dev + opam-2.2) to the base images

kit-ty-kate:
To make it easy to test opam master regularly it would be really handy to have an `opam-dev` binary around in the docker images.

### DIFF
--- a/src-opam/dockerfile_opam.ml
+++ b/src-opam/dockerfile_opam.ml
@@ -110,13 +110,14 @@ let header ?win10_revision ?arch ?maintainer ?img ?tag d =
   @@ shell
 
 (* Apk based Dockerfile *)
-let apk_opam2 ?(labels=[]) ?arch ~hash_opam_2_0 ~hash_opam_2_1 distro () =
+let apk_opam2 ?(labels=[]) ?arch ~hash_opam_2_0 ~hash_opam_2_1 ~hash_opam_master distro () =
   let img, tag = D.base_distro_tag ?arch distro in
   header ?arch distro @@ label (("distro_style", "apk") :: labels)
   @@ Linux.Apk.install "build-base bzip2 git tar curl ca-certificates openssl"
   @@ Linux.Git.init ()
   @@ install_opam_from_source ~add_default_link:false ~branch:"2.0" ~hash:hash_opam_2_0 ()
   @@ install_opam_from_source ~add_default_link:false ~enable_0install_solver:true ~branch:"2.1" ~hash:hash_opam_2_1 ()
+  @@ install_opam_from_source ~add_default_link:false ~enable_0install_solver:true ~branch:"master" ~hash:hash_opam_master ()
   @@ run "strip /usr/local/bin/opam*"
   @@ from ~tag img
   @@ Linux.Apk.add_repository ~tag:"edge" "https://dl-cdn.alpinelinux.org/alpine/edge/main"
@@ -124,6 +125,7 @@ let apk_opam2 ?(labels=[]) ?arch ~hash_opam_2_0 ~hash_opam_2_1 distro () =
   @@ Linux.Apk.add_repository ~tag:"testing" "https://dl-cdn.alpinelinux.org/alpine/edge/testing"
   @@ copy ~from:"0" ~src:["/usr/local/bin/opam-2.0"] ~dst:"/usr/bin/opam-2.0" ()
   @@ copy ~from:"0" ~src:["/usr/local/bin/opam-2.1"] ~dst:"/usr/bin/opam-2.1" ()
+  @@ copy ~from:"0" ~src:["/usr/local/bin/opam-master"] ~dst:"/usr/bin/opam-dev" ()
   @@ run "ln /usr/bin/opam-2.0 /usr/bin/opam"
   @@ Linux.Apk.dev_packages ()
   @@ Linux.Apk.add_user ~uid:1000 ~gid:1000 ~sudo:true "opam"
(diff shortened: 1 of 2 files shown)


dra27:
A thought - perhaps put the master build as `/usr/bin/opam-2.2` and, while the branch doesn't exist and also hard-link `/usr/bin/opam-dev` to it. Then when opam actually branches 2.2, things which were assuming `opam-2.2` carry on working (that binary switches to tracking the 2.2 branch) and either `/usr/bin/opam-dev` is a direct build of master or (more likely) we immediately have `/usr/bin/opam-2.3` and the same dance for `opam-dev`?

kit-ty-kate:
my reasoning is that `opam-dev` will always be the latest dev version, clients won’t have to change every time there is a new version. This binary will be always useful to have around.

If the opam-2.2 binary is opam-master, it makes it more annoying to update as we’d have to remember to update docker-base-images at branch time, especially as it will (even briefly) become 2.3.0~alpha~dev around that time. It wouldn’t make much sense.

I’d much rather have opam-2.2 only when it branches so people know it’s more stable to use.

dra27:
My reasoning was kinda the opposite 🙂 With the possibly sole exception of actually wanting to test whether opam master has broken something, most users would want to switch to a master branch because they want a feature in advance - i.e. they _want_ opam 2.2 right now but it's not released yet, so when we branch 2.2 that workflow carries on working. Having `opam-2.2` facilitates more workflows, rather than fewer. We (CI-side) should be ensuring that docker-base-images is updates at branch time already - even automatically at some point.

dra27:
(Note that I'm also suggesting that `opam-dev` is always the dev version, too)

kit-ty-kate:
Then let’s settle on a middle ground: Why not have both? We can have both:
- opam-2.2: currently pointer to master, for people who want features early but don’t want to break in the future
- and opam-dev: for people who are willing to be broken in the future and always want the latest.

dra27:
We appear to have been arguing in agreement because your last comment is what I thought I was proposing 🙂

kit-ty-kate:
ah indeed, I missed the `and also hard-link /usr/bin/opam-dev to it` part. My bad...

src-opam/dockerfile_opam.ml
```diff
@@ -192,42 +201,48 @@ let yum_opam2 ?(labels= []) ?arch ~yum_workaround ~enable_powertools ~hash_opam_
...
+  @@ copy ~from:"0" ~src:["/usr/local/bin/opam-master"] ~dst:"/usr/bin/opam-2.2" () (* TODO: Change when opam 2.2 is branched *)
+  @@ copy ~from:"0" ~src:["/usr/local/bin/opam-master"] ~dst:"/usr/bin/opam-dev" ()
   @@ run "ln /usr/bin/opam-2.0 /usr/bin/opam"
   @@ Linux.Pacman.dev_packages ()
   @@ Linux.Pacman.add_user ~uid:1000 ~sudo:true "opam"
   @@ install_bubblewrap_wrappers @@ Linux.Git.init ()
 
 (* Cygwin based Dockerfile *)
-let cygwin_opam2 ?win10_revision ?(labels=[]) ?arch ~hash_opam_2_0 ~hash_opam_2_1 distro () =
+let cygwin_opam2 ?win10_revision ?(labels=[]) ?arch ~hash_opam_2_0 ~hash_opam_2_1 ~hash_opam_master distro () =
```
avsm, line 245: It's getting to the point now where these args should be turned into a k/v pair :-)

avsm (approved):

src-opam/dockerfile_opam.ml
```diff
@@ -109,40 +109,90 @@ let header ?win10_revision ?arch ?maintainer ?img ?tag d =
...
   @@ from ~tag img
   @@ Linux.Apk.add_repository ~tag:"edge" "https://dl-cdn.alpinelinux.org/alpine/edge/main"
   @@ Linux.Apk.add_repository ~tag:"edgecommunity" "https://dl-cdn.alpinelinux.org/alpine/edge/community"
   @@ Linux.Apk.add_repository ~tag:"testing" "https://dl-cdn.alpinelinux.org/alpine/edge/testing"
-  @@ copy ~from:"0" ~src:["/usr/local/bin/opam-2.0"] ~dst:"/usr/bin/opam-2.0" ()
-  @@ copy ~from:"0" ~src:["/usr/local/bin/opam-2.1"] ~dst:"/usr/bin/opam-2.1" ()
+  @@ List.fold_left (fun acc {branch; public_names; _} ->
+      List.fold_left (fun acc public_name ->
+        acc @@ copy ~from:"0" ~src:["/usr/local/bin/opam-"^branch] ~dst:("/usr/bin/opam-"^public_name) ()
+      ) acc public_names
```
dra27, line 170: Wants some further tidying to eliminate the `assert false`, but this should shave 18M off the images (I don't think docker naturally de-dups?)
```suggestion
      match public_names with
      | [] -> assert false
      | main::aliases ->
          let main = "/usr/bin/opam-"^main in
          let acc = acc @@ copy ~from:"0" ~src:["/usr/local/bin/opam-"^branch] ~dst:main () in
          List.fold_left (fun acc public_name ->
            acc @@ run "ln %s /usr/bin/opam-%s" main public_name
          ) acc aliases
```

dra27 (commented):
Go to go once the test-run on base-image builder finishes. I don't think Docker de-dups images, so having both `opam-2.2` and `opam-dev` as copies of the same binaries is wasting 18M - suggestion to hard-link the additional names instead.

src-opam/dockerfile_opam.ml
kit-ty-kate: Done

kit-ty-kate:
Ready to go now.

src-opam/dockerfile_opam.ml
```diff
@@ -109,41 +109,95 @@ let header ?win10_revision ?arch ?maintainer ?img ?tag d =
...
+      enable_0install_solver = true;
+      public_name = "opam-2.1";
+      aliases = [];
+    };
+    {
+      branch = "master";
+      hash = opam_master_hash;
+      enable_0install_solver = true;
+      public_name = "opam-2.2"; (* TODO: Change when opam 2.2 is branched *)
+      aliases = ["opam-dev"];
```
dra27, line 152: FWIW, but it's good to go without changing it further 🙂
```suggestion
      public_name = "opam-dev";
      aliases = ["opam-2.2"]; (* TODO: Remove/update when opam 2.2 is branched *)
```

PR #79, opened 2022-01-05
Add overlays repo on Windows

dra27:
This is WIP - opening with a trial version to test the fix.


dra27:
@avsm - I think these two functions were for obi and aren't being used/tested anywhere. Up to you whether it's worth putting this in, therefore (I only landed on updating them because the Windows changes had previously been made by @MisterDA!)

MisterDA:
This is becoming interesting again, also because opam-repository-mingw is starting to be too old. It's odd that this PR wasn't merged sooner. I'm tempted to test this and add the ocaml/opam-repository to Windows images too.

dra27:
This didn't need to be merged for the base images - the repo is added there

PR #84, opened 2022-01-20
Update for OCaml 5.00

dra27:
Only amd64 at the moment and not msvc

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,7 @@
 unreleased
 ----------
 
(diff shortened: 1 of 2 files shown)


dra27:
CI likely to fail until https://github.com/ocurrent/ocaml-version/pull/44 is merged and released.

dra27:
Updated:
- CentOS/Oracle Linux 7 and earlier won't ever support OCaml 5.00.0 (GCC is too old)
- Express the conditions slightly differently to avoid the need for newer ocaml-version (for now)

PR #85, opened 2022-01-24
More efficient dependencies

dra27:
At present we build OCaml _three times_ in for each opam base image. We also occasionally build bubblewrap and then install it anyway. This is actively causing a problem on Fedora 35 where the building from source is failing (network access, or something) but we don't need it anyway.

This PR amends the Dockerfiles to build the "cold" compiler for opam's master branch and then share that with the other branches required. This has become more critical since we can use 4.13 for all of them and so avoid the sigaltstack patch mechanism on newer distros.

While here, I've refactored the aliasing mechanism for distros to eliminate a few `assert false` uses. The change should be semantically invisible for Linux (except for the intention change of return type for `resolve_alias`), but I actively changed the Windows Server 2022 tags to get rid of the ltsc2022 part (I think we made a mistake with that, @MisterDA).

### DIFF
--- a/src-opam/dockerfile_opam.ml
+++ b/src-opam/dockerfile_opam.ml
@@ -71,7 +71,7 @@ let bubblewrap_and_dev_packages distro =
   | `Yum -> Linux.RPM.dev_packages
   | `Zypper -> Linux.Zypper.dev_packages
   | `Pacman -> Linux.Pacman.dev_packages
-  | _ -> assert false
+  | `Cygwin | `Windows -> assert false
   in
   match D.bubblewrap_version distro with
   | Some version when version >= bubblewrap_minimum ->
--- a/src-opam/dockerfile_opam.ml
+++ b/src-opam/dockerfile_opam.ml
@@ -30,25 +30,24 @@ let personality ?arch distro =
 
 let run_as_opam fmt = Linux.run_as_user "opam" fmt
 
-let install_opam_from_source ?(add_default_link=true) ?(prefix= "/usr/local") ?(enable_0install_solver=false) ~branch ~hash () =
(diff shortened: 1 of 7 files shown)


MisterDA:
Not too sure about the LTSC2022 part.
There are currently no `21H1` nor `21H2` tags for [ServerCore](https://mcr.microsoft.com/v2/windows/servercore/tags/list), so the should either return `` `Deprecated`` or something else, but not `` `Active``. This is also the case for `win10_latest_image`. There's only a `ltsc2022` tag and `win10-21h1-preview` which should be deleted some time in the future.

dra27:
OK - the removal of ltsc2022 from _our_ tags should be OK, though? We might need to do something richer - possibly as a future PR? At present, we have Windows 10 21H2 and Windows 11 21H2 - Server 2022 is simply an alias of Windows 11 21H2.

src-opam/dockerfile_opam.ml
```diff
@@ -50,14 +50,36 @@ let install_opam_from_source_cygwin ?(add_default_link=true) ?(prefix= "/usr/loc
...
+    run "rm -rf %s bubblewrap-%s" file rel
+
+let bubblewrap_and_dev_packages distro =
+  let dev_packages = match D.package_manager distro with
+  | `Apk -> Linux.Apk.dev_packages
+  | `Apt -> Linux.Apt.dev_packages
+  | `Yum -> Linux.RPM.dev_packages
+  | `Zypper -> Linux.Zypper.dev_packages
+  | `Pacman -> Linux.Pacman.dev_packages
+  | _ -> assert false
```
tmcgilchrist, line 68: Maybe as ?
```
| `Cygwin | `Windows -> assert false
```

tmcgilchrist (commented):
The reduction in assert falses' is pleasing.

PR #86, opened 2022-01-24
Install devtoolset-7 for RHEL 7 derivatives

dra27:
RHEL 7 and clones (CentOS, Oracle Linux, etc.) have GCC 4.8 which is too old for multicore. It is possible to get GCC 7 via [RedHat Software Collections](https://www.softwarecollections.org/en/) which this PR starts to enable.

Accessing the commands needs some trickery via the `scl` command (`scl enable devtoolset-7 bash` gives you a shell with the appropriate tweaks - I haven't looked in enough detail to see if you can persist it).

There'd then be the question of whether to have _all_ CentOS 7 and Oracle Linux 7 images using GCC 7 (probably bad - better for the images for the older compilers to use the stock GCC, however old it is) or to build a separate opam image specifically for use by OCaml 5.00.0+ (more work, more space).

Anyway, this is what I'd done so far.

### DIFF
--- a/src-opam/dockerfile_opam.ml
+++ b/src-opam/dockerfile_opam.ml
@@ -221,10 +221,16 @@ let yum_opam2 ?(labels= []) ?arch ~yum_workaround ~enable_powertools ~opam_hashe
       @@ Linux.RPM.install "yum-plugin-ovl"
     else empty
   in
+  let repo_or_update, extra =
+    match distro with
+    | `CentOS `V7 -> Linux.RPM.install "centos-release-scl", Some "devtoolset-7"
+    | `OracleLinux `V7 -> Linux.RPM.install "oracle-softwarecollection-release-el7", Some "devtoolset-7"
+    | _ -> Linux.RPM.update, None
+  in
   header ?arch distro @@ label (("distro_style", "rpm") :: labels)
   @@ run "yum --version || dnf install -y yum"
   @@ workaround
-  @@ Linux.RPM.update
+  @@ repo_or_update
   @@ Linux.RPM.dev_packages ~extra:"which tar curl xz libcap-devel openssl" ()
   @@ Linux.Git.init ()
   @@ install_bubblewrap_from_source ()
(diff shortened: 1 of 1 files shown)


avsm:
See https://github.com/ocurrent/ocaml-dockerfile/pull/31 -- the difficult bit is persisting the `scl`. If we had that, it would also unlock better Rust support in our CI for RHEL7

edwintorok:
There is also devtoolset-11 which is newer, and works on CentOS 7 too (I'm using it).
Here is the change I had to make to ocaml.spec to enable it (albeit for 4.13.x, not 5.0 yet):
```
diff --git a/ocaml.spec b/ocaml.spec
index 08c48e5..8e409ae 100644
--- a/ocaml.spec
+++ b/ocaml.spec
@@ -64,7 +64,7 @@ Source0:        https://repo.citrite.net/ctx-local-contrib/xs-opam/ocaml-%{versi

 BuildRequires:  make
 BuildRequires:  git
-BuildRequires:  gcc
+BuildRequires: devtoolset-11-gcc devtoolset-11-binutils
 BuildRequires:  autoconf
 BuildRequires:  binutils-devel
 BuildRequires:  ncurses-devel
@@ -76,7 +76,7 @@ BuildRequires:  /usr/bin/annocheck

 # ocamlopt runs gcc to link binaries.  Because Fedora includes
 # hardening flags automatically, redhat-rpm-config is also required.
-Requires:       gcc
+Requires:       devtoolset-11-gcc
 Requires:       redhat-rpm-config

 # Because we pass -c flag to ocaml-find-requires (to avoid circular
@@ -174,6 +174,8 @@ unset MAKEFLAGS
 make=make
 %endif

+source /opt/rh/devtoolset-11/enable
+
 # Don't use %%configure macro because it sets --build, --host which
 # breaks some incorrect assumptions made by OCaml's configure.ac
 #
```

devtoolset doesn't modify the PATH on purpose, each package has to opt-in by sourcing the appropriate file.

This is not persisted in 'native_c_compiler' output of 'ocamlc -config' because that still refers to 'gcc' (although using older GCC to compile stubs "works", but obviously won't help if the headers themselves want newer compiler).

It might be possible to "persist" the devtoolset compiler by overriding CC (and related) flags instead of just extending the PATH, another alternative is to drop a script into `/etc/profile.d` in the container that sources this script.

This would be similar to how we handle 'eval $(opam env)' in RPM builds:
```
mkdir -p %{buildroot}/etc/profile.d
mkdir -p %{buildroot}%{_opamroot}
echo 'export OPAMROOT=%{_opamroot}' > %{buildroot}/etc/profile.d/opam.sh
echo 'export OPAMROOT=%{_opamroot}' > %{buildroot}/etc/profile.d/opam.sh
echo 'eval `opam config env`' >> %{buildroot}/etc/profile.d/opam.sh```

dra27:
> This is not persisted in 'native_c_compiler' output of 'ocamlc -config' because that still refers to 'gcc' (although using older GCC to compile stubs "works", but obviously won't help if the headers themselves want newer compiler).

This will matter for 5.0 - the `_Atomic` is required by pretty much every header.

PR #87, opened 2022-01-30
Remove opam-repository from the dockerfiles created by dockerfile-opam

kit-ty-kate:
See https://github.com/ocurrent/docker-base-images/issues/99

To me having opam-repository is a straight downside in every regard. For the few specialized CI systems that require it for some reason they can clone it by hand and rely on the docker cache to offset the time cost.

This is a breaking change and should be announced on Discuss before merging of course.


avsm:
I still can't form an opinion about this PR without knowing which users it will break (outside of the CI systems above).   Perhaps posting on discuss.ocaml.org about the proposed change would be worthwhile first, @kit-ty-kate ?

benmandrew:
@kit-ty-kate any updates on this?

PR #88, opened 2022-02-01
CentOS 8 is EOL since 31/12/2021

kit-ty-kate:
Fixes https://github.com/ocurrent/docker-base-images/issues/157

### DIFF
--- a/src-opam/dockerfile_distro.ml
+++ b/src-opam/dockerfile_distro.ml
@@ -224,10 +224,9 @@ let distro_status (d:t) : status = match d with
   | `Alpine `V3_15 -> `Active `Tier1
   | `Alpine `Latest -> `Alias (`Alpine `V3_15)
   | `Archlinux `Latest -> `Active `Tier3
-  | `CentOS `V8 -> `Active `Tier2
   | `CentOS `V7 -> `Active `Tier3
-  | `CentOS `V6 -> `Deprecated
-  | `CentOS `Latest -> `Alias (`CentOS `V8)
+  | `CentOS (`V6 | `V8) -> `Deprecated
+  | `CentOS `Latest -> `Alias (`CentOS `V7)
   | `Debian (`V7|`V8|`V9) -> `Deprecated
   | `Debian `V10 -> `Active `Tier2
   | `Debian `V11 -> `Active `Tier1


kit-ty-kate:
https://www.centos.org/centos-linux-eol/

dra27:
I haven't looked at the detail, but shouldn't we rather be switching to CentOS Stream?

dra27:
It feels very strange to me for the centos-latest tag to go backwards. There are some possibilities (we could do several of them):
- Use the Quay Container Registry for the base image (`quay.io/centos/centos:stream8`)
- Add RedHat Base Images (Docker Hub - [redhat/ubi8](https://hub.docker.com/r/redhat/ubi8) or [redhat/ubi8-minimal](https://hub.docker.com/r/redhat/ubi8-minimal)if we can build with that)
- Add Rocky Linux (and possibly even point the centos-latest tag to that?!)
- Stop updating the centos-latest tag

dra27:
In terms of right now, because we certainly need to get CentOS 8 out of the distros list, given that it can't even update, would it be OK to restore the alias of `` `CentOS `Latest`` to point to `` `V8`` but _remove_ `` `CentOS `Latest` from both `Dockerfile_distro.distros` and `Dockerfile_distro.latest_distros`. Certainly on the base image builder that stops it from pushing `ocaml/opam:centos-*` images for now which means that they're simply out-of-date as opposed to suddenly jumping back to CentOS 7. Then we can decide with less pressure on what to do (if anything, eventually) about CentOS/RHEL/Rocky

kit-ty-kate:
I’m not sure leaving `` `CentOS `Latest`` to point to `` `V8`` is the right thing to do. For example other distributions such as Debian have their `` `Latest`` pointing to the "latest stable" (11 currently) instead of "latest latest" (11 or sid).

To me 7 is their currently available "latest stable" version at the moment.

dra27:
Heh, heh - other distributions don't stop support for version _N_ **before** version _N-1_ 🙂 I guess you're right - it's seriously weird, but so's having `ocaml/opam:centos-ocaml-*` being old and out-of-date. I guess I'll have to hold my nose!

tmcgilchrist:
> Add Rocky Linux 
Seems like the most like for like replacement for CentOS.

dra27:
Nose held! 🙂

PR #90, opened 2022-03-01
Revert #36 (temporary workaround for Archlinux when Docker installation < 20.10.4)

kit-ty-kate:
It looks like the workers got updated at some point since


dra27 (approved):
Seems to be working, thanks!

PR #92, opened 2022-03-22
Unset the jobs variable from the opam config (faulty opam 2.0 behaviour)

kit-ty-kate:
opam 2.0 by default stores the number of cores of the current machine. While not too much of a big deal for users, this is extremely bad for the base images where the number stored will be around 72 but these can be used on any other machines.

This might help fighting the IO overload on the cluster.

### DIFF
--- a/src-opam/dockerfile_opam.ml
+++ b/src-opam/dockerfile_opam.ml
@@ -404,6 +404,7 @@ let all_ocaml_compilers hub_id arch distro =
     @@ sandbox
     @@ run "opam init -k git -a /home/opam/opam-repository --bare%s"
          (if os_family = `Windows then " --disable-sandboxing" else "")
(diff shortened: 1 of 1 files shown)


tmcgilchrist (approved):

dra27:
The problem doing this in the base is that it breaks 2.0 - opam 2.1 recomputes the `jobs` variable if it's missing, but opam 2.0 just assumes "1". I think this wants to be in workflows as part of the root upgrade step?

benmandrew:
@kit-ty-kate can this be merged now that the opam PR is merged?

kit-ty-kate:
@benmandrew the opam PR only fixes things when upgrading to opam 2.2. David’s comment above is still very much the main blocker, so until the images are upgraded to a more recent opam by default this PR can’t be merged.

PR #96, opened 2022-05-23
Use opam2 branch when pulling from opam-repository-mingw

MisterDA:
There seem to be no users of `all_ocaml_compilers` in all the pipelines I've examined, which explains why this had not been caught before.


MisterDA:
cc @dra27

PR #99, opened 2022-06-28
Support BuildKit 1.4 syntax (heredocs, `--link`)

MisterDA:
I just saw that [BuildKit 1.4 syntax](https://github.com/moby/buildkit/blob/master/frontend/dockerfile/docs/syntax.md#here-documents) has support for here-documents in `RUN` and `COPY` instructions. I've added support for `COPY`, and used it in the small shell-scripts we output.
I've also added a new `link` flag to the `COPY` and `ADD` instructions.
The "breaking change" is that the generated Dockerfile from `gen_opam2_distro` now uses BuildKit syntax.
BuildKit syntax is only supported with Linux containers.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,7 @@ unreleased
 ----------
 
 - Support BuildKit 1.4 syntax of here-documents in `COPY` instructions. (@MisterDA #99)
+- Support BuildKit 1.4 `--link` flag in `ADD` and `COPY` instructions. (@MisterDA #99)
 
 v8.0.0 2022-07-27 Sydney
 ------------------------
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,8 @@
 unreleased
 ----------
 
+- Support BuildKit 1.4 syntax of here-documents in `COPY` instructions. (@MisterDA #99)
+
 v8.0.0 2022-07-27 Sydney
 ------------------------
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,8 @@
+unreleased
+----------
+
 v8.0.0 2022-07-27 Sydney
------------------
+------------------------
 
 - Deprecate Ubuntu 21.10 (@tmcgilchrist #104)
 - Various LCU Updates (@mtelvers #103 #98 #95 #93 #91 #89 #83)
--- a/src/dockerfile.ml
+++ b/src/dockerfile.ml
@@ -35,6 +35,16 @@ type parser_directive =
   [ `Syntax of string | `Escape of char ]
   [@@deriving sexp]
 
+type heredoc = {
+  here_document: string;
+  word: string;
+  delimiter: string;
+  strip: bool } [@@deriving sexp]
+
+type heredocs_to_dest =
+  [`Chown of string option] * heredoc list * string
+  [@@deriving sexp]
+
 type line =
   [ `ParserDirective of parser_directive
   | `Comment of string
@@ -46,6 +56,7 @@ type line =
   | `Env of (string * string) list
   | `Add of sources_to_dest
   | `Copy of sources_to_dest
+  | `Copy_heredoc of heredocs_to_dest
   | `Entrypoint of shell_or_exec
   | `Shell of string list
   | `Volume of string list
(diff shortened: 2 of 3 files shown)


src-opam/dockerfile_opam.ml
```diff
@@ -81,23 +81,32 @@ let bubblewrap_and_dev_packages distro =
...
+{|	#!/bin/sh
+	cp ~/.opamrc-sandbox ~/.opamrc
+	echo --- opam sandboxing enabled|} in
+  let sandbox_disable = heredoc ~strip
+{|	#!/bin/sh
+	cp ~/.opamrc-nosandbox ~/.opamrc
+	echo --- opam sandboxing disabled|} in
+  (* Disable bubblewrap *)
+  copy_heredoc ~chown:"opam" ~src:[opamrc_nosandbox] ~dst:"/home/opam/.opamrc-nosandbox" () @@
+  copy_heredoc ~chown:"opam" ~src:[sandbox_disable] ~dst:"/home/opam/opam-sandbox-disable" () @@
```
maiste, line 104: IMHO, this part of the syntax is harder to read than the original version.  The `{| |}` introduces complexity by mixing ocaml and bash syntax instead of the `run "cmd"`.

MisterDA: Thank you for the review Etienne. I'd say the complexity is having to follow the file line-by-line through multiple `RUN` and `echo >` calls and that the code is more "readable", or at least that the file isn't being split and stands out in the OCaml as well as in the generated Dockerfile.
Since you don't share that opinion, but also that I think we can agree it's a minor change without much technical advantages, I'm going to split the commit using heredocs (even though it was the feature that motivated this change) from this PR so that maybe it can move forward.

src/dockerfile.mli
```diff
@@ -138,7 +156,7 @@ val env : (string * string) list -> t
   instructions. This is functionally equivalent to prefixing a shell
   command with [<key>=<value>]. *)
 
-val add : ?chown:string -> ?from:string -> src:string list -> dst:string -> unit -> t
+val add : ?link:bool -> ?chown:string -> ?from:string -> src:string list -> dst:string -> unit -> t
 (** [add ?from ~src ~dst ()] copies new files, directories or remote file URLs
   from [src] and adds them to the filesystem of the container at the
   [dst] path.
```
maiste, line 162: Maybe here add the  `?link` with a small explanation about what it does?

maiste (approved):
Thanks for the split in two PRs! As you said, the syntax is an opinionated subject, so I'll comply with the majority :+1: 
The PR  LGTM! You can add the `?link` remark if you want to or not (as it also appears in the buildkit documentation).
However, I don't have the merge right on `ocaml-dockerfile` so I can't merge it....

src/dockerfile.mli
tmcgilchrist: Agree, please document what `?link:bool` does.

CHANGES.md
```diff
@@ -39,6 +39,8 @@ v8.0.0 unreleased
 - Expose `Dockerfile_distro.win10_docker_base_image` and
   `Dockerfile_distro.win10_base_tag` to get the Windows container base
   image and tags. (@MisterDA #63)
+- Support BuildKit 1.4 syntax of here-documents in `COPY` instructions. (@MisterDA #99)
+- Support BuildKit 1.4 `--link` flag in `ADD` and `COPY` instructions. (@MisterDA #99)
```
tmcgilchrist, line 43: This will need to move to `0.8.1 unreleased` depending on how impactful the breaking change is.

src/dockerfile.ml
```diff
@@ -129,17 +140,33 @@ let optional name = function
   | None -> []
   | Some value -> [sprintf "%s=%s" name value]
 
+let optional_flag name = function
+  | Some true -> [name]
+  | _ -> []
```
tmcgilchrist, line 145: Please exhaustively match here.

tmcgilchrist (approved):
Please address the 2 remaining comments before merge.

MisterDA:
> Please address the 2 remaining comments before merge.

Done, thanks!

tmcgilchrist (approved):

PR #100, opened 2022-07-06
Added riscv64

mtelvers:


src-opam/dockerfile_distro.ml
```diff
@@ -360,7 +360,7 @@ let distro_arches ov (d:t) =
   | `Debian `V10, ov when OV.(compare Releases.v4_02_0 ov) = -1 -> [ `I386; `X86_64; `Aarch64; `Aarch32]
   | `Debian `V9, ov when OV.(compare Releases.v4_03_0 ov) = -1 -> [ `I386; `X86_64; `Aarch64; `Aarch32 ]
   | `Alpine (`V3_6 | `V3_7 | `V3_8 | `V3_9 | `V3_10 | `V3_11 | `V3_12 | `V3_13 | `V3_14 | `V3_15), ov when OV.(compare Releases.v4_05_0 ov) = -1 -> [ `X86_64; `Aarch64 ]
-  | `Ubuntu (`V18_04|`V20_04|`V20_10|`V21_04|`V21_10|`V22_04), ov when OV.(compare Releases.v4_05_0 ov) = -1  -> [ `X86_64; `Aarch64; `Ppc64le ]
+  | `Ubuntu (`V18_04|`V20_04|`V20_10|`V21_04|`V21_10|`V22_04), ov when OV.(compare Releases.v4_05_0 ov) = -1  -> [ `X86_64; `Aarch64; `Ppc64le; `Riscv64 ]
```
dra27, line 363: This function is already duplicating logic from `Ocaml_version.Since`... it'd be better to refactor it completely not to be doing that, but in the meantime:
```suggestion
  | `Ubuntu `V18_04, ov when OV.(compare Releases.v4_05_0 ov) = -1  -> [ `X86_64; `Aarch64; `Ppc64le; `Riscv64 ]
  | `Ubuntu (`V20_04|`V20_10|`V21_04|`V21_10|`V22_04), ov when OV.(compare Releases.v4_05_0 ov) = -1  ->
    let base = [ `X86_64; `Aarch64; `Ppc64le ] in
    if OV.(compare Releases.v4_11_0 ov) <= 0 then
      `Riscv64 :: base
    else
      base
```
This simultaneously addresses:
- No riscv64 image for 18.04
- riscv64 only for 4.11+

dra27 (commented):
Temporary change - it'd be better to be have `distro_arches` reflect the architectures available on that distribution (in Docker) and then use `Ocaml_version.Since` to filter those architectures based on the specific ocaml version, as that would remove the duplicated logic (for example, the manual encoding of Debian for ppc64le and s390 being 4.03+)

PR #105, opened 2022-07-25
Use BuildKit heredocs

MisterDA:
- [ ] #99 

>> IMHO, this part of the syntax is harder to read than the original version.  The `{| |}` introduces complexity by mixing ocaml and >> bash syntax instead of the `run "cmd"`.
>>
>> _Originally posted by @maiste in https://github.com/avsm/ocaml-dockerfile/pull/99#discussion_r927634254_
>
> Thank you for the review Etienne. I'd say the complexity is having to follow the file line-by-line through multiple `RUN` and `echo >` calls and that the code is more "readable", or at least that the file isn't being split and stands out in the OCaml as well as in the generated Dockerfile.
> Since you don't share that opinion, but also that I think we can agree it's a minor change without much technical advantages, I'm going to split the commit using heredocs (even though it was the feature that motivated this change) from this PR so that maybe it can move forward.
>
> _Originally posted by @MisterDA in https://github.com/avsm/ocaml-dockerfile/pull/99#discussion_r928958169_


src-opam/dockerfile_opam.ml
```diff
@@ -81,23 +81,32 @@ let bubblewrap_and_dev_packages distro =
...
-  run "echo 'wrap-remove-commands: []' >> ~/.opamrc-nosandbox" @@
-  run "echo 'required-tools: []' >> ~/.opamrc-nosandbox" @@
-  run "echo '#!/bin/sh' > /home/opam/opam-sandbox-disable" @@
-  run "echo 'cp ~/.opamrc-nosandbox ~/.opamrc' >> /home/opam/opam-sandbox-disable" @@
-  run "echo 'echo --- opam sandboxing disabled' >> /home/opam/opam-sandbox-disable" @@
+  let strip = true in
+  let opamrc_sandbox = heredoc ~strip
+{|	wrap-build-commands: ["%%{hooks}%%/sandbox.sh" "build"]
+	wrap-install-commands: ["%%{hooks}%%/sandbox.sh" "install"]
+	wrap-remove-commands: ["%%{hooks}%%/sandbox.sh" "remove"]|} in
```
tmcgilchrist, line 88: I would add the `|} in` on a newline wherever you use it. Makes it easier to read and show what is multiline string and what is the actual command.

tmcgilchrist (approved):
Small formatting change, otherwise it looks good.

src-opam/dockerfile_opam.ml
MisterDA: The problem is that the end delimiter has to go on a newline. The formatting function will add `"\n%s" t.delimiter` (usually `\nEOF`), so adding a new line here would render 

```
COPY --chown=opam <<-EOF /home/opam/opam-sandbox-disable
	#!/bin/sh
	cp ~/.opamrc-nosandbox ~/.opamrc
	echo --- opam sandboxing disabled

EOF
```

instead of
```
COPY --chown=opam <<-EOF /home/opam/opam-sandbox-disable
	#!/bin/sh
	cp ~/.opamrc-nosandbox ~/.opamrc
	echo --- opam sandboxing disabled
EOF
```

I'm not sure checking whether the heredoc ends with a newline or not is good, because the heredoc should be preserved verbatim. What to you think?

tmcgilchrist: The more compact version makes sense.

PR #106, opened 2022-07-27
Some misc. chores

MisterDA:
- gen opam files with Dune, bump to Dune 3.0;
- bump to ocaml 4.08 and remove deps on result and rresult;
- apply ocamlformat.0.24.1;
- wrap libraries.


.ocamlformat
```diff
@@ -0,0 +1,3 @@
+profile = conventional
+version = 0.23.0
+ocaml-version = 4.11
```
tmcgilchrist, line 3: Use version 0.24.1

src-cmd/utils.ml
```diff
@@ -0,0 +1,17 @@
+let ( >>= ) = Result.bind
+
+let rec iter fn l =
+  match l with hd :: tl -> fn hd >>= fun () -> iter fn tl | [] -> Ok ()
+
+let map fn l =
```
tmcgilchrist, line 6: Generate an interface file for this.

tmcgilchrist (approved):

.ocamlformat
```diff
@@ -0,0 +1,3 @@
+profile = conventional
+version = 0.24.1
+ocaml-version = 4.11
```
tmcgilchrist, line 3: Is there a strong reason to move the OCaml version this far forward?
https://ocaml.org/p/dockerfile/8.0.0 supports back to 4.02.3

MisterDA: We can remove the dependency on rresult, use new functions from the standard library. I think there's a case for moving slowly to new versions. An informal rule is to support the last three OCaml versions. I can go back to 4.10 for now.

MisterDA: the CI doesn't currently test before 4.07.

tmcgilchrist (approved):

PR #108, opened 2022-09-01
Adapt Windows opam-repository to new git safe.directory setting

MisterDA:
Mitigations added for CVE-2022-24765 cause error messages as the owner
of the git checkout won't be correct. It simply doesn't matter on a CI
system, so just add the directory to safe.directory to ignore it.

cc @dra27

### DIFF
--- a/src-opam/dockerfile_windows.ml
+++ b/src-opam/dockerfile_windows.ml
@@ -148,7 +148,10 @@ module Cygwin = struct
   module Git = struct
     let init ?(cyg=default) ?(name="Docker") ?(email="docker@example.com") () =
       env ["HOME", cyg.root ^ {|\home\opam|}]
-      @@ run_sh ~cyg "git config --global user.email '%s' && git config --global user.name '%s' && git config --system core.longpaths true" email name
+      @@ run_sh ~cyg "git config --global user.email '%s' && \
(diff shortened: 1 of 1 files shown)


src-opam/dockerfile_windows.ml
```diff
@@ -228,6 +231,9 @@ module Winget = struct
 
   module Git = struct
     let init ?(name="Docker") ?(email="docker@example.com") () =
-      run "git config --global user.email %S && git config --global user.name %S && git config --system core.longpaths true" email name
+      run "git config --global user.email %S && \
+           git config --global user.name %S && \
+           git config --global core.longpaths true && \
```
MisterDA, line 236: Note that the user doesn't have the permissions to set directives system-wide, so I also fixed `core.longpaths` to `--global`. By default, cygwin's git is used.

dra27: Is it possible to do this earlier in a step as ContainerAdministrator?

MisterDA: I re-tested and must have misread, the `--system` switch for `core.longpaths` worked as expected, both for Cygwin's git and Git for Windows. Reverting that change.
As a side note, the user is always Container Administrator. I haven't created another user.

PR #114, opened 2022-10-07
Build Cygwin (& OCaml for Windows) in a sep multistage image, then copy

MisterDA:
This last commit changes the Windows images to install Cygwin and OCaml for Windows in a separate multistage image. The idea is that the installation never leaves `C:\cygwin64` and doesn't change the environment apart from adding cygwin's `/bin` to the `PATH`, but we're doing that manually anyway, so it should be safe to copy `C:\cygwin64`.
If we ever get some form of BuildKit for Windows, then Winget and Cygwin retrieval and installation will be done in parallel!
I think we gain more readability by closing in behaviours of WinGet and Cygwin installs, separation of concerns from the main image, and overall better understanding of the end image. It's possible that we also may gain better sharing when building the images. Also, we don't have to remove the cache from the build image.

Includes:
- #112 
- #113

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,10 @@
 unreleased
 ----------
 
+- Refactor to be able to install Cygwin and OCaml for Windows in a
+  separate multistage build image, then copy. Docker Engine 20.10.18
+  for Windows is currently buggy and doesn't allow it.
+  (@MisterDA #114)
 - Support `ARG` Dockerfile instruction (@MisterDA #116 #117)
 - Bump to OCaml 4.08 and remove dependencies on result and rresult (@MisterDA #106)
 - Wrap libraries:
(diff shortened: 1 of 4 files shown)


MisterDA:
Converted to draft because I haven't tested it yet ;)

MisterDA:
> Merge consecutive RUN instructions in Windows images
>
> Although this hurts the readability of the end imge, this little
    optimization allows us to merge intermediary images—which means faster
    uploading and fetching of Windows images.
>
> For debugging this can simply be removed.

Turns out changes to `PATH` and the environment variables are not propagated to later commands if they're "crunched" (as cmd parses everything in one go?) so calling `crunch` on the whole image doesn't work.

MisterDA:
Deployment has succeeded.

PR #117, opened 2022-10-12
Support Dockerfile ARG instruction

### LINKED ISSUE #116: Add support for ARG
https://docs.docker.com/engine/reference/builder/#arg

MisterDA:
Should `arg` rather have as signature:

```ocaml
val arg : string -> ?default:string -> unit -> t (* my current choice *)
val arg : (string * string option) -> t
```

cc @bn-d.
Close #116.


src/dockerfile.ml
```diff
@@ -211,6 +219,7 @@ let cmd fmt = ksprintf (fun b -> [ `Cmd (`Shell b) ]) fmt
 let cmd_exec cmds : t = [ `Cmd (`Exec cmds) ]
 let expose_port p : t = [ `Expose [ p ] ]
 let expose_ports p : t = [ `Expose p ]
+let arg a ?default () : t = [ `Arg (a, default) ]
```
maiste, line 222: Why not:
```suggestion
let arg ?default a : t = [ `Arg (a, default) ]
```
This way, you don't have to include the `unit` at the end.

maiste (approved):
LGTM :+1: 
I suggested another syntax. WDYT?

bn-d:
I would second @maiste suggestion. Otherwise, looks great to me! Thanks for the swift response.

src/dockerfile.ml
maiste: As it's a label argument, not sure it's an issue :)

PR #118, opened 2022-10-13
Initial support for MacOS Monterey using homebrew

tmcgilchrist:


src-opam/distro.ml
```diff
@@ -187,6 +187,8 @@ let win10_kb_number_to_lcu (v : win10_release) kb =
   | lcu -> Some (v, lcu)
   | exception Not_found -> None
 
+type macos_all = [ `Latest | `Macos_monterey ]
```
kit-ty-kate, line 190: I think using their version number is much easier to maintain and remember which version they are
```suggestion
type macos_all = [ `Latest | `V12 ]
```

src-opam/distro.ml
```diff
@@ -562,6 +571,7 @@ let distro_status (d : t) : status =
         `Deprecated
     | `Cygwin v -> win10_docker_status `ServerCore v
     | `Windows (_, v) -> win10_docker_status `Windows v
+    | `Macos _ -> `Active `Tier3 (** TODO What does this Tier level mean? Matches Windows Tier3 *)
```
kit-ty-kate, line 574: Tier level is used to get a grouping on a set of distributions so they can have different properties:

* Tier 1 will have all the different variants and architectures and should be tested thoroughly
* Tier 2 are less supported but still important distributions to test as extra
* Tier 3 are new or badly supported distributions in opam-repository. Those are usually disabled in ocaml-ci (otherwise it would fail too often) and enabled in opam-repo-ci for example (to ensure its support can get better over time)

src-opam/distro.ml
```diff
@@ -868,6 +879,7 @@ let tag_of_distro (d : t) =
   | `Cygwin v -> "cygwin-" ^ win10_release_to_string v
   | `Windows (`Mingw, v) -> "windows-mingw-" ^ win10_release_to_string v
   | `Windows (`Msvc, v) -> "windows-msvc-" ^ win10_release_to_string v
+  | `Macos (`Latest | `Macos_monterey) -> "macos-homebrew-12.6"
```
kit-ty-kate, line 882: I don't think we should store patches versions for macos as there are usually no significant difference and it changes too often.
```suggestion
  | `Macos (`Latest | `Macos_monterey) -> "macos-homebrew-12"
```

src-opam/distro.ml
```diff
@@ -1264,7 +1281,8 @@ let base_distro_tag ?win10_revision ?(arch = `X86_64) d =
   | `Windows (_, v) ->
       win10_base_tag ?win10_revision `Windows
         (v : win10_release :> [> win10_release ])
-
+  | `Macos `Macos_monterey -> ("macos/monterey", "12.6") (* TODO There is no docker image for this yet! *)
```
kit-ty-kate, line 1284:
```suggestion
  | `Macos `Macos_monterey -> ("macos/monterey", "12") (* TODO There is no docker image for this yet! *)
```

src-opam/dockerfile_opam.ml
```diff
@@ -490,7 +494,7 @@ let all_ocaml_compilers hub_id arch distro =
     @@ env [ ("OPAMYES", "1") ]
     @@
     match os_family with
-    | `Linux | `Cygwin -> cmd "bash"
+    | `Linux | `Cygwin | `Macos -> cmd "bash" (* TODO Should we use zsh since that is the default now in MacOS? *)
```
kit-ty-kate, line 497: I don't think it would change much given this setting is only used for the dockerfile itself not really the underlying setup (which can't really be tested in CI)

src-opam/dockerfile_opam.ml
```diff
@@ -532,6 +536,7 @@ let separate_ocaml_compilers hub_id arch distro =
            let sandbox =
              match os_family with
              | `Linux -> run "opam-sandbox-disable"
+             | `Macos -> run "opam-sandbox-disable" (* TODO Not sure about this? *)
              | `Windows | `Cygwin -> empty
```
kit-ty-kate, line 540: macos does not have any sandbox so it should absolutely have the opam sandbox
```suggestion
             | `Windows | `Cygwin | `Macos -> empty
```

avsm (changes requested):
There's no explanation as to what this PR is actually for.  There's no upstream support for native macOS in Docker, so what is this actually doing?

tmcgilchrist:
@avsm This is relying on the obuilder / ocluster support for running a docker like environment. We are currently using this to submit opam-repo-ci jobs to test macos-homebrew (Monterey). My intention is to add this support into `ocaml-ci` and move the workarounds in opam-repo-ci back into this repository. 

There is a point to be made about no official native docker support on MacOS and this not being usable outside of obuilder. I could move it into another module with clear documentation saying only for obuilder.
Do you think this is being used by outside of our CI use case? 

https://github.com/ocurrent/ocluster/pull/193 
https://github.com/ocurrent/obuilder/pull/122

src-opam/distro.ml
tmcgilchrist: The plan is to support macOS Monterey on x86_64 with a small range of OCaml versions (4.14 and 5.0). That sounds like a Tier2 to me?

In future we would look at supporting Apple Silicon as that is clearly the future for macOS developer tools, but that is planned to come after macOS x86_64 and Windows support is rolled out. We will get more benefit for the effort from hitting macOS and Windows, over an additional architecture that is partially tested on Linux ARM64.

kit-ty-kate: Either Tier2 or Tier3 is fine for me (the macos support in opam-repo-ci isn't flowless yet)

avsm:
Thanks @tmcgilchrist.  On reflection, I don't think it's appropriate to put this into ocaml-dockerfile, as there isn't a macOS port.  It's probably a good idea to expose something slightly higher level for the purposes of OCaml-ci -- you will run across the same problem for a future *BSD support there.  Proposals welcome for what that library might look like: it would be nice to be able to express a CI strategy that is portable across obuilder _and_ Docker _and_ other mechanisms in the future as we develop them.

tmcgilchrist:
I agree @avsm. Closing this in favour of an obuilder spec based library.

PR #119, opened 2022-10-13
Add alpine.3.16

raphael-proust:
Alpine 3.16 has been released a few months ago. This PR adds support for it.

### DIFF
--- a/src-opam/distro.ml
+++ b/src-opam/distro.ml
@@ -207,7 +207,8 @@ type distro =
     | `V3_12
     | `V3_13
     | `V3_14
-    | `V3_15 ]
+    | `V3_15
+    | `V3_16 ]
   | `Archlinux of [ `Latest ]
   | `CentOS of [ `V6 | `V7 | `V8 ]
   | `Debian of [ `V11 | `V10 | `V9 | `V8 | `V7 | `Testing | `Unstable ]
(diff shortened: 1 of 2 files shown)


src-opam/distro.ml
```diff
@@ -542,6 +545,7 @@ let distro_status (d : t) : status =
         `Deprecated
     | `Alpine `V3_14 -> `Active `Tier2
     | `Alpine `V3_15 -> `Active `Tier1
+    | `Alpine `V3_16 -> `Active `Tier1
```
raphael-proust, line 548: Note sure about this. I read the documentation on tiers but I still don't know if that's the correct tier to set for the alpine versions.

kit-ty-kate: 3.15 should probably be downgraded to Tier2. Otherwise this would create all the variant images for both 3.15 and 3.16

avsm (approved):
lgtm once the comment is addressed. Let's get the CI running 3.16 as soon as possible as its been out for 4 months already.

On a broader subject, it might be worth pushing `alpine-edge` images as well (without putting them in CI, but building ocaml on it) so that we can catch bleeding edge issues in Alpine. It's become a pretty important distro for OCaml users these days, and we do already test debian-unstable in a similar fashion.

tmcgilchrist:
Antonin’s pr adding it to the base image builder
https://github.com/ocurrent/docker-base-images/pull/190

PR #120, opened 2022-10-13
Build bubblewrap 0.6.2 from source for older distributions

MisterDA:
I don't think we're dealing with these old distros anymore but we may benefit from the bug fixes of bubblewrap. No breakage reported in its changelog.

cc @raphael-proust


kit-ty-kate (approved):
sure, why not

avsm (approved):
Fine by me.

avsm:
(please push a CHANGES entry for this)

PR #121, opened 2022-10-17
Support `STOPSIGNAL` instruction

MisterDA:
https://docs.docker.com/engine/reference/builder/#stopsignal

Is exposing `stopsignal_int` and `stopsignal_name` better than ``stopsignal [`Int of int | `Name of string]``? It's probably more consistent this way.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,7 @@
 unreleased
 ----------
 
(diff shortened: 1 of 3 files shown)


avsm:
Since this is constructing a Dockerfile, I'd prefer not to expose the integer one at all, and always generate the SIG* version.  It's too errorprone to refer directly to the kernel's syscall table when a perfectly descriptive alternative is available.

PR #122, opened 2022-10-17
Support `HEALTHCHECK` instruction

MisterDA:
https://docs.docker.com/engine/reference/builder/#healthcheck

### DIFF
--- a/src/dockerfile.mli
+++ b/src/dockerfile.mli
@@ -322,6 +322,49 @@ val label : (string * string) list -> t
 
   To view an image’s labels, use the [docker inspect] command. *)
 
+val healthcheck :
+  ?interval:string ->
+  ?timeout:string ->
+  ?start_period:string ->
+  ?retries:int ->
+  ('a, unit, string, t) format4 ->
+  'a
+(** [healthcheck cmd] checks container health by running a command
+  inside the container. See {!cmd} for additional details.
+
+  @param interval The health check will first run [interval] seconds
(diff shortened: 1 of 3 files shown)


src/dockerfile.mli
```diff
@@ -322,6 +322,49 @@ val label : (string * string) list -> t
...
+  ?timeout:string ->
+  ?start_period:string ->
+  ?retries:int ->
+  ('a, unit, string, t) format4 ->
+  'a
+(** [healthcheck cmd] checks container health by running a command
+  inside the container. See {!cmd} for additional details.
+
+  @param interval The health check will first run [interval] seconds
+    after the container is started, and then again interval seconds
```
avsm, line 336:
```suggestion
    after the container is started, and then again [interval] seconds
```

src/dockerfile.mli
```diff
@@ -322,6 +322,49 @@ val label : (string * string) list -> t
...
+    after each previous check completes.
+
+  @param timeout If a single run of the check takes longer than
+    [timeout] seconds then the check is considered to have failed.
+
+  @param retries It takes [retries] consecutive failures of the health
+    check for the container to be considered _unhealthy_.
```
avsm, line 343:
```suggestion
    check for the container to be considered {i unhealthy}.
```

avsm (approved):
lgtm apart from minor comments. This was a feature added to Docker by @talex5, iirc!

PR #123, opened 2022-10-17
Remove opam-2.0 def, defaults to opam-2.1

MisterDA:
Defaulting to opam-2.1 removes the step of converting the opamroot to the new layout. Should the definition need to be kept, we could switch the fields of `opam_hashes` to `string option` to trigger the build of the corresponding opam version.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,7 @@
 unreleased
 ----------
 
+- Remove opam 2.0 definition, defaults to opam 2.1. (@MisterDA, #123)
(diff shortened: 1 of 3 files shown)


avsm (changes requested):
Feel free to change the default to opam 2.1 as it's now the recommended release, but why on earth remove opam 2.0.x while it's still maintained for security fixes (and distributed with e.g. Ubuntu 20.04 LTS).  We keep testing it until it's deprecated and has no users -- that's the point of CI.

Issue #17, opened 2019-11-26
Fedora don't use YUM anymore, should migrate to DNF


XVilka:
For many releases DNF is the default package manager, maybe it makes sense to update it from Yum to DNF? It will require dropping some older versions though.

avsm:
PR welcome for this.

PR #127, opened 2022-11-07
Switch from yum to DNF for Fedora, CentOS and OracleLinux

MisterDA:
Tentative fix for #17.

### DIFF
--- a/src-opam/distro.ml
+++ b/src-opam/distro.ml
@@ -1093,16 +1093,16 @@ let latest_tag_of_distro (t : t) =
   tag_of_distro latest
 
 type package_manager =
-  [ `Apt | `Yum | `Apk | `Zypper | `Pacman | `Cygwin | `Windows ]
+  [ `Apt | `Dnf | `Apk | `Zypper | `Pacman | `Cygwin | `Windows ]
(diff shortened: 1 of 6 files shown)


kit-ty-kate:
opam-depext (for 2.0) does not support dnf so if you really want to see this through you'll need to send a fix and release opam-depext first

src-opam/distro.ml
```diff
@@ -1093,16 +1093,16 @@ let latest_tag_of_distro (t : t) =
...
 [@@deriving sexp]
 
 let package_manager (t : t) =
   match t with
   | `Ubuntu _ -> `Apt
   | `Debian _ -> `Apt
-  | `CentOS _ -> `Yum
-  | `Fedora _ -> `Yum
-  | `OracleLinux _ -> `Yum
+  | `CentOS _ -> `Dnf
```
edwintorok, line 1103: Not ALL CentOS supports DNF. CentOS 7 (which is what I'm currently using the docker containers for) does not have DNF, but just Yum, so might be useful to retain 'Yum' as a package manager and just add DNF.
(Although Yum is still present as an alias in newer versions of Fedora, although unsure for how long)

PR #130, opened 2022-11-29
Alpine: don't use libexecinfo-dev

MisterDA:


MisterDA:
For even more context, symbols from libexecinfo (`backtrace`, `backtrace_symbols`, `backtrace_symbols_fd`) were used for debugging in ocaml-multicore. That's why they're in the source of 4.12+domains+effects, but were removed for OCaml 5.0 alpha, in https://github.com/ocaml/ocaml/commit/905d17c95ddadfc919d34ef156f656bde8939b55#diff-5dbfd2b27b3f223dd26f1b82a2db605021d953d98f628f37737b9da56c491b60.
They're still used for a specific test in OCaml's test suite, but it's fine not to depend on this lib in Alpine anymore.

PR #132, opened 2022-12-16
Include the ocaml-beta-repository in the images

kit-ty-kate:
`opam init --reinit` needed to update the opam state needs network access for nothing otherwise.


MisterDA (changes requested):
lgtm, if that's still needed would you please run `dune fmt` before merging?

PR #133, opened 2022-12-17
debian:10 does not provide ppc64 and s390x images anymore

kit-ty-kate:
See https://hub.docker.com/_/debian/tags?page=1&name=10


dra27:
Alternate in #134 which restores pcc64 and s390x for Debian 10, rather than removing it. Thanks for debugging and looking at these failures!

PR #134, opened 2022-12-17
Tweak the exotic architecture docker references

dra27:
Alternate for #133 - it's not that they're not pushed anymore, they're not multiarch on older versions.


dra27:
Last commit tested locally on an arm64 box, and it does indeed fix a long-standing problem with the Dockerfiles - for the 32-bit (arm and i386) they were not specifying the platform+personality for the second (main) part of the build so this only worked when we used a non-multiarch image. This if course didn't affect ppl64/s390x (modulo the problem with Debian 10) because there's only one image in the multiarch which they can run.

dra27:
The tags are now pushing from docker-base-images, but once that's correct I'll merge this and update them again.

avsm:
Be a little careful on the multi-platform builders, as Docker had some unexpected caching behaviour in older releases that may still linger. Specifically, `--platform` may sometimes be ignored: https://github.com/docker/cli/issues/2590

This should now be fixed, but also need to look at https://github.com/docker/buildx/issues/844 to make sure it isn't an issue.  We should be entirely manifest-based now, so I hope it isn't a problem.

dra27:
Ta: let's definitely leave testing the last commit on the base image builder until 5.0 is completely deployed, then! The code needs reviewing regardless.

dra27:
Fully tested on docker-base-images now, and it seems to be working correctly

MisterDA:
Merged as of 60eb31dbea462ca993fa971ae62dc00443027eac with a `dune fmt` fix, thanks!

Issue #137, opened 2023-01-24
mount type cache support


edwintorok:
TL;DR:
* add `?mounts:(Mount.t list)` to `Dockerfile.run` and support in `Dockerfile.crunch`
* add a simple Mount.t: mount type, list of options (or string map of options)
* introduce ?mount_cache  in dockerfile-opam's package managers (default off), which would mount `/var/cache/yum` with an appropriately computed cache id (perhaps overridable with `?cache_id`)
* provide some helpers for an 'opam' command with appropriate caches mounted (`~/.cache` for `dune`, and a symlinked `~/.opam/download-cache/{md5,sha256,sha512} to .cache`), usable by both 'opam install' and 'opam monorepo pull' for downstream containers to use
* would this be useful for you in https://github.com/ocurrent/docker-base-images too? Anything in particular I should be aware of to make it useful there?

I've got some very early experimental code to add support for `--mount=type=cache` on RUN lines that works both with Docker BuildKit and Podman 4.x (see https://docs.docker.com/engine/reference/builder/#run---mounttypecache).

For now it is as a layer on top of `dockerfile` and `dockerfile-opam`, but I'd like to contribute at least some of the changes back to this library. In particular `Dockerfile.crunch` needs to know about the mounts, because the correct way to crunch this:
```
RUN --mount=type=cache,target=/var/cache sudo yum install -y foo
RUN --mount=type=cache,target=/home/opam/.cache opam install bar
```

is
```
RUN --mount=type=cache,target=/var/cache --mount=type=cache,target=/home/opam/.cache sudo yum install -y foo && opam install bar
```

(i.e. the mounts need to stay grouped together at the beginning)

And once that is in place then the various Package managers in dockerfile-opam could be taught to take a `?use_cache` parameter to enable a package download cache (shared among all dockerfiles, not just the current one, with some care to use a proper cache ID per OS/architecture as needed!), and skip the 'clean all' at the end.
Which is beneficial even on fast networks (the package mirrors are sometimes very slow, and especially on CentOS/Fedora just refreshing the mirror/package metadata can take significantly longer than downloading the package).

Also a cache mount type is very useful for downstream container builds that do 'opam install' or 'dune build' or 'opam depext',
or 'opam monorepo pull', all of which can be cached.

[...]

PR #139, opened 2023-02-02
Support `RUN` command arguments (mounts, network, security)

MisterDA:
This is a first attempt at supporting `RUN` command arguments. I've added types for mounts and networks which are in BuildKit, and security, which is not yet in BuildKit. I've added an escape hatch to `RUN` as the `?args:string list` parameter for later additions (and podman compat?). The security and network types, being relatively simple now, get their own parameters to `RUN`, but the mount values are serialized to string first.
The `crunch` function will not crunch shell scripts if their network or security differ, specified or not.
cc #137

### DIFF
--- a/src/dockerfile.ml
+++ b/src/dockerfile.ml
@@ -106,12 +106,14 @@ type healthcheck_options = {
 type healthcheck = [ `Cmd of healthcheck_options * shell_or_exec | `None ]
 [@@deriving sexp]
 
+type network = [ `Default | `None | `Host ] [@@deriving sexp]
+
 type line =
   [ `ParserDirective of parser_directive
   | `Comment of string
   | `From of from
   | `Maintainer of string
-  | `Run of mount list * shell_or_exec
+  | `Run of mount list * network option * shell_or_exec
   | `Cmd of shell_or_exec
   | `Expose of int list
   | `Arg of string * string option
@@ -148,12 +150,15 @@ let crunch l =
   let pack l =
     let rec aux acc = function
       | [] -> acc
-      | `Run (m, `Shell a) :: `Run (m', `Shell b) :: tl ->
-          aux (`Run (merge m m', `Shells [ a; b ]) :: acc) tl
-      | `Run (m, `Shells a) :: `Run (m', `Shell b) :: tl ->
-          aux (`Run (merge m m', `Shells (a @ [ b ])) :: acc) tl
-      | `Run (m, `Shells a) :: `Run (m', `Shells b) :: tl ->
-          aux (`Run (merge m m', `Shells (a @ b)) :: acc) tl
+      | `Run (m, n, `Shell a) :: `Run (m', n', `Shell b) :: tl ->
+          if n <> n' then invalid_arg "crunch: at least two networks differ.";
+          aux (`Run (merge m m', n, `Shells [ a; b ]) :: acc) tl
+      | `Run (m, n, `Shells a) :: `Run (m', n', `Shell b) :: tl ->
+          if n <> n' then invalid_arg "crunch: at least two networks differ.";
+          aux (`Run (merge m m', n, `Shells (a @ [ b ])) :: acc) tl
+      | `Run (m, n, `Shells a) :: `Run (m', n', `Shells b) :: tl ->
+          if n <> n' then invalid_arg "crunch: at least two networks differ.";
+          aux (`Run (merge m m', n, `Shells (a @ b)) :: acc) tl
       | hd :: tl -> aux (hd :: acc) tl
     in
     List.rev (aux [] l)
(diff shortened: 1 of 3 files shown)


MisterDA:
@edwintorok I had not read your proposal carefully enough, do you already have code to achieve this? What do you think of this proposal?

edwintorok:
My code is currently here https://github.com/ocurrent/ocaml-dockerfile/compare/master...edwintorok:ocaml-dockerfile:caching. I'm not entirely happy with how the API on that looks, and still evolving it as I use it.

src/dockerfile.ml
```diff
@@ -243,6 +295,65 @@ let parser_directive pd : t = [ `ParserDirective pd ]
...
+        @ (match sharing with
+          | None -> []
+          | Some `Shared -> [ "sharing=shared" ]
+          | Some `Private -> [ "sharing=private" ]
+          | Some `Locked -> [ "sharing=locked" ])
+        @ optional "from" from @ optional "source" source
+        @ optional_int_octal "mode" mode
+        @ optional_int "uid" uid @ optional_int "gid" gid)
+  | `Tmpfs { target; size } ->
+      sprintf "--mount=type=tmpfs,target=%s,size=%d" target size
```
edwintorok, line 384: size is only supported by Docker. I don't think Podman supports size, just target, so perhaps size could be made optional?

edwintorok:
Crunching mounts: I'm undecided what the best way forward is. Crunching commands with different mounts, especially with locking=shared might result in the build holding the lock for way longer than intended (small individual commands may be faster here even if they create additional layers).
Perhaps there should be an option to crunch to specify whether RUN lines with different mounts can be crunched together or not (in my changes so far I defaulted to no, and in your changes you defaulted to yes, so if we make it configurable it'll work for both of us).

edwintorok:
> @edwintorok I had not read your proposal carefully enough, do you already have code to achieve this? What do you think of this proposal?

I think I can rebase my changes on top of yours, in some areas they are more complete than mine.
Sorry if I made you duplicate some of what I already done (perhaps I should've opened a draft PR sooner). 

I have a few more things on my branch (caching support in opam package manager, but that builds on top of this, and some doc comments, and a convenient value for the correct parsing_directive to use).

MisterDA:
Crunching is already super fragile: is assumes a Unix shell-like syntax. I think we had an issue for something like this but I can't remember where it is. Anyway, I don't use crunch too much in Windows code, it behaves poorly with line continuations, `&&`, between cmd, powershell, Cygwin's shell...
It's doable to have a configurable switch to commands with different mounts, but I fear crunch would do too much. Perhaps it should error instead?

> I have a few more things on my branch (caching support in opam package manager, but that builds on top of this, and some doc comments, and a convenient value for the correct parsing_directive to use).

Cool!

edwintorok:
as long as the mounts are identical I think crunch should accept it (and be careful so we end up just with one set of mounts), otherwise either eject it or silently skip over it. Rejecting sounds better because it wouldn't break user expectations.

Yes crunching is already quite fragile if you e.g. 'cd' in one RUN line then the next one will run in a different directory. That may or may not be what the user wanted. It is useful to be able to run on small portions of the dockerfile, but when mounts are used I don't think that running it on the entire one would be useful.

src/dockerfile.ml
MisterDA: Done, thanks!

MisterDA:
I've simplified the code, I think. I've added your `buildkit_syntax` value too.
`run` and `run_exec` take well typed arguments.
`crunch` will raise if it tries to merge lines using different networks or security policies. It'll merge mounts if they are strictly identical.

edwintorok:
Thanks, I have some additional code to use this, I'll rebase it on top of your latest branch.

edwintorok:
https://github.com/ocurrent/ocaml-dockerfile/compare/run-args...edwintorok:ocaml-dockerfile:caching-run-args

There are some commits there that you may want to cherry-pick into this PR:
2dcbb8eb724c66d8765a0e25cb13ad01b4fc4908 Dockerfile: escape spaces in mount args
45dac5c5bfe09d9b05199b69053c8a3ce3b00d06 Dockerfile.mli: more documentation on mounts
fe66805f073d4228819a5edff547f46d078b91ab Dockerfile: refer to {!buildkit_syntax}

In particular escaping is important because 'space' is otherwise a separator between mount args, but it may also show up inside mount args (e.g. a cache id, or a directory path, though in general on Linux paths don't have spaces that is not a reason to not implement escaping).

I've also written a test that generates a single dockerfile with all opam containers and builds them in parallel, but I'll open a separate PR with that (it is in my branch above).

edwintorok (approved):

MisterDA:
> I've also written a test that generates a single dockerfile with all opam containers and builds them in parallel, but I'll open a separate PR with that (it is in my branch above), and the rest of the changes (to use caching for apt/yum/opam commands/etc.)

The caching part is super cool, that would be a great feature IMO.
As a matter of fact, there's already a test outputting all the build instructions to give us a point of reference when they change; it's just not in this repo. See [`ocurrent/docker-base-images/builds.expected`](https://github.com/ocurrent/docker-base-images/blob/master/builds.expected).

Thank you for the other commits, I've cherry-picked them (rebasing with the format fix, which changed the commit id).

I'm all in for escapes, so I've included your patch.

tmcgilchrist:
Additional tests would be most welcome, especially on the escaping spaces.

tmcgilchrist:
This looks great. Thank you.

PR #140, opened 2023-02-03
Build and install opam master from source in Windows images

MisterDA:
Currently testing this.
I'm only building opam master, using mingw-w64 (64 bits). Two aliases are available, `opam-dev` and `opam-2.2`.

### DIFF
--- a/src-opam/opam.ml
+++ b/src-opam/opam.ml
@@ -438,7 +438,7 @@ let windows_opam2 ?win10_revision ?winget ?(labels = []) ?arch ~opam_hashes
          install_from_release
            ~extra:
              ("git" :: "patch" :: "mingw64-x86_64-gcc-g++"
-            :: "mingw64-i686-gcc-g++" :: mingw_packages ())
+            :: "mingw64-i686-gcc-g++" :: mingw_packages)
            ())
     @@ install_opams_windows opam_master_hash opam_branches
   in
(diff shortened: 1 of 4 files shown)


src-opam/opam.ml
```diff
@@ -213,11 +258,22 @@ let install_opams ?prefix opam_master_hash opam_branches =
...
-         @@ install_opam_from_source ?add_default_link ?prefix
-              ?enable_0install_solver ~branch ~hash ())
+         @@ install_opam_from_source ~add_default_link:false ?prefix
+              ~enable_0install_solver ~branch ~hash ())
+       empty opam_branches
+
+let install_opams_windows ?cyg ?prefix opam_master_hash opam_branches =
+  Windows.Cygwin.Git.init ?cyg ~repos:[ "/tmp/opam-sources" ] ()
+  @@ Windows.Cygwin.run_sh ?cyg
+       "git clone https://github.com/MisterDA/opam /tmp/opam && cd /tmp/opam \
```
kit-ty-kate, line 269:
```suggestion
       "git clone https://github.com/ocaml/opam /tmp/opam && cd /tmp/opam \
```

kit-ty-kate (changes requested):
It looks like the HEAD commit of your master branch has been merged upstream so this shouldn’t be necessary anymore

src-opam/opam.ml
MisterDA: Yes, sure, that was for testing :)

src-opam/opam.ml
```diff
@@ -47,6 +53,38 @@ let install_opam_from_source ?(add_default_link = true) ?(prefix = "/usr/local")
...
+           Windows.Cygwin.run_sh ?cyg "cd /tmp/opam-build-%s && make lib-pkg"
+             branch
+          else empty)
+       @@ Windows.Cygwin.run_sh ?cyg
+            "cd /tmp/opam-build-%s && ./configure --enable-cold-check%s && make"
+            branch
+            (if enable_0install_solver then " --with-0install-solver" else ""))
+  @@ Windows.Cygwin.run_sh ?cyg
+       "cd /tmp/opam-build-%s && mkdir -p %s/bin && cp /tmp/opam-build-%s/opam \
+        %s/bin/opam-%s && chmod a+x %s/bin/opam-%s"
```
MisterDA, line 84: I tried removing the build directory, but `rm` complained about an non-empty dir. Presumably, some file inside of it couldn't be removed (permission issue?). That's not too bad since I'm using a separate build stage.

kit-ty-kate (approved):

PR #147, opened 2023-03-03
Optimize Linux package installation

MisterDA:
Crunch update / upgrade / install as much as possible. Clean caches.

- [ ] https://github.com/ocurrent/docker-base-images/pull/212

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,7 @@
 unreleased
 ----------
 
+- Optimize Linux package install. (@MisterDA #147)
 - Various LCU Updates (@mtelvers #144 #136 #135)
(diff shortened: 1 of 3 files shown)


MisterDA:
This seems to be working as expected!

kit-ty-kate:
nope

kit-ty-kate:
I'm reverting this. The system package manager's cache should always be present, otherwise opam depext does not work anymore.

avsm:
Agreed. I think the old opam-depext equivalent of `opam depext -u` no longer exists in opam 2.1, which would have not encountered this problem (as it would update the package manager db before doing the querying).  It's not in 2.1 as that's too slow for the integrated depext.

PR #149, opened 2023-03-06
Install system packages required by OCaml 5.1+ in the ocaml stage 

MisterDA:
Different releases of OCaml require different system libraries: libX11 isn't needed since 4.09, when the Graphics library was moved from the core distribution to a separate package. On the opposite side, OCaml 5.1 can optionally use libzstd for compressing marshalled data. Install these packages accordingly in the ocaml image.

### DIFF
--- a/src-opam/linux.ml
+++ b/src-opam/linux.ml
@@ -45,6 +45,11 @@ module RPM = struct
        libX11-devel which m4 diffutils findutils%s"
       (match extra with None -> "" | Some x -> " " ^ x)
 
+  let ocaml_depexts v =
+    if Ocaml_version.compare v Ocaml_version.Releases.v5_1_0 >= 0 then
+      Some "zstd"
+    else None
+
   let add_user ?uid ?gid ?(sudo = false) username =
     let uid = match uid with Some u -> sprintf "-u %d " u | None -> "" in
     let gid = match gid with Some g -> sprintf "-g %d " g | None -> "" in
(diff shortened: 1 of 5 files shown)


MisterDA:
Now based on top of #153.

kit-ty-kate (commented):
That sounds good in principle but I'm not sure why the `dev_packages` function wasn't changed instead of creating a new function that users have to call manually. Is there a downside to adding a new mandatory parameter to the `dev_packages` function? Is this just to allow optional dependencies to be decided by the user (if so I think the function name could be more explicit about that). I would personally be more in favour of simply adding a new mandatory parameter to `dev_packages` and enforce a set of packages in an opinionated way.

MisterDA:
hmm, passing the switch in an option would work too, I'll try that. 
Thanks for the idea!

src-opam/linux.ml
```diff
@@ -39,11 +39,20 @@ module RPM = struct
...
+        install
+          "sudo passwd bzip2 unzip patch rsync nano gcc-c++ git tar curl xz \
+           which m4 diffutils findutils%s"
+          extra
+    | Some v ->
+        if Ocaml_version.compare v Ocaml_version.Releases.v4_08 <= 0 then
+          install "libX11-devel%s" extra
+        else if Ocaml_version.compare v Ocaml_version.Releases.v5_1_0 >= 0 then
+          install "zstd%s" extra
+        else empty
```
kit-ty-kate, line 55: i was more thinking of something like this
```suggestion
    install
      "sudo passwd bzip2 unzip patch rsync nano gcc-c++ git tar curl xz \
       which m4 diffutils findutils%s%s"
      extra
      (if Ocaml_version.compare v Ocaml_version.Releases.v4_08 <= 0
       then "libX11-devel"
       else if Ocaml_version.compare v Ocaml_version.Releases.v5_1_0 >= 0
       then "zstd"
       else ""
```

src-opam/linux.ml
```diff
@@ -121,10 +131,15 @@ module Apk = struct
 
   let dev_packages ?extra () =
     install
-      "build-base patch tar ca-certificates git rsync curl sudo bash \
-       libx11-dev nano coreutils xz ncurses-dev%s"
+      "build-base patch tar ca-certificates git rsync curl sudo bash nano \
+       coreutils xz ncurses-dev libX11-dev%s"
```
kit-ty-kate, line 135: On Alpine the name of packages are case sensitive
```suggestion
       coreutils xz ncurses-dev libx11-dev%s"
```
I would also suggest to keep the same ording as before to limit the diff and avoid such issues elsewhere

PR #153, opened 2023-03-10
Optimise and fix Linux packages installation

MisterDA:
Try to reorder calls to update the package lists, upgrade the current installed packages, and install new packages to minimize network calls. Some fixes along the way. Make sure that installing packages updates the package lists first.
Most package managers don't have a package download cache enabled, but pacman does. Cleaning the pacman dl cache recovers ~112M (with libx11 as a dependency, YMMV: depends whether new packages have been published since the build of the base image).
Some package managers require package lists to be available before upgrading/installing packages (apt), others will automatically download them. Be conservative, and keep the package lists in all cases.
opam can't download the package lists if the package manager doesn't do it automatically, and will fail the `opam depext` install if so.
There's no particular need to crunch the lines into simpler layers, since the images are later squashed.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,7 @@
 unreleased
 ----------
 
+- Optimize and fix Linux package install (@MisterDA #153)
 - Switch to ocaml-opam/opam-repository-mingw#sunset for Windows images. (@MisterDA #152)
 - Use DockerHub user risvc64/ubuntu. (@MisterDA, #150)
 - Various LCU Updates (@mtelvers #144 #136 #135)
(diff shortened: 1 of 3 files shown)


kit-ty-kate:
I’m not sure why I was asked for a review if it’s going to be merged the next day (weekend) without any input.

There is a problem on distributions using yum: `yum clean all` removes the package database.

MisterDA:
> I’m not sure why I was asked for a review if it’s going to be merged the next day (weekend) without any input.

You're right, I should have left this waiting longer, sorry.

> There is a problem on distributions using yum: `yum clean all` removes the package database.

(This wasn't changed in this PR)
That's right, but yum and dnf have the courtesy of fetching the package database if it doesn't exist, so `opam depext` will not fail. `yum clean packages` is probably more efficient. The package database for Fedora is ~110Mb, so there's a case for including it in the image even if it's heavy, to avoid re-downloading it every time.

kit-ty-kate:
> (This wasn't changed in this PR)

it was because the database was redownloaded at the next `yum install` just after.

PR #159, opened 2023-04-06
Make docstring indentation consistent + some docs rewording

benmandrew:
Docstrings are inconsistently indented, so picked one.

Is there a reason we don't use `ocamlformat`?

### DIFF
--- a/src-opam/distro.mli
+++ b/src-opam/distro.mli
@@ -146,7 +146,7 @@ type distro =
   | `Cygwin of win10_release
   | `Windows of [ `Mingw | `Msvc ] * win10_release ]
 [@@deriving sexp]
-(** Supported Docker container distributions distributions. *)
+(** Supported Docker container distributions without aliases. *)
 
 type t =
   [ `Alpine of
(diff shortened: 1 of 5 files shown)


src-opam/distro.mli
```diff
@@ -223,7 +223,7 @@ type t =
   | `Cygwin of win_all
   | `Windows of [ `Mingw | `Msvc ] * win_all ]
 [@@deriving sexp]
-(** Supported Docker container distributions. *)
+(** Supported Docker container distributions with aliases such as [Latest]. *)
```
MisterDA, line 226:
```suggestion
(** Supported Docker container distributions with aliases such as [`Latest]. *)
```

MisterDA:
> Docstrings are inconsistently indented, so picked one.
> 
> Is there a reason we don't use `ocamlformat`?

We do use ocamlformat. Maybe it just doesn't process doc blocks.
https://ocaml.ci.dev/github/ocurrent/ocaml-dockerfile/commit/b6706540a5917b1eda1dabf0930d5ea59c14a2f5/variant/%28lint-fmt%29

PR #162, opened 2023-04-12
Refactor and improve Windows builds

MisterDA:
opam was being build with mingw even in the msvc images. Separate mingw and msvc builds into two functions. Separate the msvc build into more build steps. Fix the origin of Install.cmd from avsm/ocaml-dockerfile to ocurrent/ocaml-dockerfile.
See diff at https://github.com/ocurrent/docker-base-images/pull/223.


MisterDA:
This requires https://github.com/ocaml/opam/pull/5514.

src-opam/windows.ml
```diff
@@ -252,13 +280,24 @@ module Cygwin = struct
...
+      [
+        "rsync";
+        "patch";
+        "diffutils";
+        "curl";
+        "make";
+        "unzip";
+        "git";
+        "m4";
+        "perl";
```
tmcgilchrist, line 299: Why does this require Perl? Dropping `vim` seems reasonable for CI images though it might surprise some users of the Windows images.

MisterDA: Usually the perl dependency comes from git. Perl is listed in the installation instructions of [OCaml for Windows](https://fdopen.github.io/opam-repository-mingw/installation/). We'll drop it once we remove OCaml for Windows.
I'm inclined to remove vim and nano altogether from the CI images, but as it may break workflows we don't know about I won't do it. I thought that starting from scratch, the Windows images didn't need them.

PR #164, opened 2023-04-24
Added Ubuntu 23.04 and Fedora 38

mtelvers:
- Ubuntu 23.04 was released 20 April 2023
- Fedora 38 was released 18 April 2023

Note that the Docker base images for Ubuntu 23.04 include a default user called `ubuntu` with UID 1000.  See this [post](https://bugs.launchpad.net/cloud-images/+bug/2005129)

The opam build assumes that UID 1000 is available and runs `adduser --uid 1000 --disabled-password --gecos '' opam`. This PR introduces a new step which runs `if getent passwd 1000; then userdel -r $(id -nu 1000); fi` to delete any user using UID 1000 before attempting to add the new user.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,7 @@
 unreleased
 ----------
 
+- Added Ubuntu 23.04 (@mtelvers)
 - Add newlines in some cases for better formatting.
   (@MisterDA #161, review by @benmandrew)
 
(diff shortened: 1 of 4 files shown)


MisterDA (approved):
I'll trust you on the user thing.
Thanks!

You just need to run `dune fmt` to fix the formatting errors. I suggest you rebase and run `dune fmt` on each commit.

PR #177, opened 2023-06-26
Adopt the OCaml Code of Conduct

rikusilvola:
The OCaml Code of Conduct can be found in [ocaml/code-of-conduct](https://github.com/ocaml/code-of-conduct) and has been discussed [in this Discourse thread](https://discuss.ocaml.org/t/ocaml-community-code-of-conduct/10494).

We propose adopting it for ocurrent/ocaml-dockerfile as well.

cc @MisterDA @tmcgilchrist

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,0 +1,16 @@
+# Code of Conduct
+
+This project has adopted the [OCaml Code of Conduct](https://github.com/ocaml/code-of-conduct/blob/main/CODE_OF_CONDUCT.md).
+
+# Enforcement
+
+This project follows the OCaml Code of Conduct [enforcement policy](https://github.com/ocaml/code-of-conduct/blob/main/CODE_OF_CONDUCT.md#enforcement).
+
+To report any violations, please contact:
+
+* Antonin Décimo <antonin [at] tarides [dot] com>
+* Anil Madhavapeddy <anil [at] recoil [dot] org>
+* David Allsopp <david.allsopp [at] metastack [dot] com>
+* Kate Deplaix <kit-ty-kate [at] outlook [dot] com>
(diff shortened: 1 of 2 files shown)


CODE_OF_CONDUCT.md
```diff
@@ -0,0 +1,13 @@
+# Code of Conduct
+
+This project has adopted the [OCaml Code of Conduct](https://github.com/ocaml/code-of-conduct/blob/main/CODE_OF_CONDUCT.md).
+
+# Enforcement
+
+This project follows the OCaml Code of Conduct [enforcement policy](https://github.com/ocaml/code-of-conduct/blob/main/CODE_OF_CONDUCT.md#enforcement).
+
+To report any violations, please contact:
```
samoht, line 9: This doesn't look like the full list of maintainers: https://github.com/ocurrent/ocaml-dockerfile/blob/master/dockerfile.opam#L9-L15

avsm:
Rather than adding this individually to every single repo, might we just have a central organisation one please? See https://github.com/geocaml/.github for an example. It makes maintaining the list of contacts more practical.

tmcgilchrist:
> Rather than adding this individually to every single repo, might we just have a central organisation one please? See https://github.com/geocaml/.github for an example. It makes maintaining the list of contacts more practical.

We've setup a default organisation code of conduct. My plan was to override that for certain projects (ocaml-version, ocaml-dockerfile, opam-repo-ci, opam-health-check) to include a wider set of names reflecting the larger number of contributors @avsm

avsm:
An organisation is intended as a place to set common standards (of maintainership, of conduct, of purpose, etc.) for the repositories contained within.  It is also a pragmatic choice, since the organisation acts as one source of truth for metadata that is otherwise spread across multiple projects.  As such, I would prefer for it to be all consistent in the organisation, and individual projects within ocurrent not singled out.

PR #203, opened 2024-04-17
Add Fedora 40

mtelvers:
Addresses https://github.com/ocaml/infrastructure/issues/113.  However, Fedora won't be released until 23rd April so do not merge yet!


MisterDA:
In the same spirit Ubuntu 24.04 is scheduled for April 25.

PR #209, opened 2024-06-20
Remove Debian 10 ppc63le and s390

mtelvers:
Ahead of the EOL of Debian 10 it seems that Debian have already removed the mirrors for POWER9 and S390X.


MisterDA (approved):
Maybe the interface isn't flexible enough to express that _pairs_ of distro and arches can be (or not) deprecated?
Curiously, why aren't we supporting ppc64le and s390x on Debian 11 and 12?

mtelvers:
We do for OCaml > 4.03.

https://github.com/ocurrent/ocaml-dockerfile/blob/7730768b0094fbe8d24dfbac3412166ba64a995d/src-opam/distro.ml#L444-L445

MisterDA:
Thanks, I was only looking at the diff and not the whole picture...

PR #210, opened 2024-06-20
Deprecate Debian 10

shonfeder:
Depian 10 LTS reaches end of life on June 30th (see https://lists.debian.org/debian-lts-announce/2024/05/msg00002.html).

This should be merged at that time, and then updates propagated through our CI services.

Pleases let me know if there is anything else I should change!

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,7 @@
 unreleased
 ----------
 
+- Deprecate Debian 10. (@shonfeder, #210)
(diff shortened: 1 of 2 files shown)


tmcgilchrist:
@shonfeder This will need to be rolled out to the various CI services (opam-ci, ocaml-ci and base image builder in particular) by changing the git submodule they include for this dependency.

PR #223, opened 2024-11-04
Fix for Fedora 41

mtelvers:
Updates the syntax of `dnf` to support DNF 5 used in Fedora 41.

DNF5 complete drops `groupinstall` where previously it has just been deprecated.


shonfeder (approved):

shonfeder:
Note that this is arguably a work around for #17

PR #227, opened 2025-04-02
Match case between FROM and AS

Firobe:
To avoid warnings of the form
```
#1 WARN: FromAsCasing: 'as' and 'FROM' keywords' casing do not match
```
generate `FROM ... AS ...` clauses instead of `FROM ... as ...`


MisterDA:
Thanks!

PR #237, opened 2025-08-11
Add Debian 13 (Trixie)

MisterDA:
[Debian 13 _"trixie"_ released](https://www.debian.org/News/2025/20250809)

Note that:

> i386 is no longer supported as a regular architecture: there is no official kernel and no Debian installer for i386 systems. The i386 architecture is now only intended to be used on a 64-bit (amd64) CPU. Users running i386 systems should not upgrade to trixie. Instead, Debian recommends either reinstalling them as amd64, where possible, or retiring the hardware.


src-opam/distro.ml
```diff
@@ -461,6 +474,10 @@ module OV = Ocaml_version
 let distro_arches ov (d : t) =
   match (resolve_alias d, ov) with
   | (`CentOS (`V6 | `V7) | `OracleLinux `V7), ov when OV.major ov >= 5 -> []
+  | `Debian `V13, ov when OV.(compare Releases.v4_03_0 ov) = -1 ->
+      [ `X86_64; `Aarch64; `Aarch32; `Ppc64le; `S390x ]
+  | `Debian `V13, ov when OV.(compare Releases.v4_02_0 ov) = -1 ->
+      [ `X86_64; `Aarch64; `Aarch32 ]
```
mtelvers, line 480: Excitingly, riscv64 is now supported.
```suggestion
  | `Debian `V13, ov when OV.(compare Releases.v4_11_0 ov) = -1 ->
      [ `I386; `X86_64; `Aarch64; `Aarch32; `Ppc64le; `S390x; `Riscv64 ]
  | `Debian `V13, ov when OV.(compare Releases.v4_03_0 ov) = -1 ->
      [ `I386; `X86_64; `Aarch64; `Aarch32; `Ppc64le; `S390x ]
  | `Debian `V13, ov when OV.(compare Releases.v4_02_0 ov) = -1 ->
      [ `I386; `X86_64; `Aarch64; `Aarch32 ]
```

PR #239, opened 2025-09-25
Track Docker Layers

fiifikcbentil:
This PR introduces a new layers function `val layers : t -> int` that approximates the number of filesystem layers a given Dockerfile (`Dockerfile.t`) will produce.
The function inspects each instruction in the Dockerfile (`Dockerfile.t`) and counts only those that are guaranteed to create a new image layer in Docker, see: https://docs.docker.com/engine/storage/drivers/#images-and-layers. These are `FROM`, `RUN`, `COPY`, and `ADD`.
Being able to estimate the layer count from `Dockerfile.t` will help inform decisions regarding image size and build performance as well as help analyze the cost of adding new instructions

### DIFF
--- a/src/dockerfile.ml
+++ b/src/dockerfile.ml
@@ -577,7 +577,6 @@ let string_of_t tl =
   Buffer.contents buf
 
 let layers (t : t) : int =
-  let t = crunch t in
   let is_layer = function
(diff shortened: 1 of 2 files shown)


MisterDA (approved):
LGTM, thanks!

PR #248, opened 2025-11-04
Packaging chores

MisterDA:


dockerfile.opam
```diff
@@ -33,7 +33,7 @@ doc: "https://ocurrent.github.io/ocaml-dockerfile/doc/dockerfile/"
 bug-reports: "https://github.com/ocurrent/ocaml-dockerfile/issues"
 depends: [
   "dune" {>= "3.0"}
-  "ocaml" {>= "4.08"}
+  "ocaml" {>= "4.14"}
```
mtelvers: Why do we want to do this?

MisterDA: I'd like to stop having to care about these old OCaml versions, and stop the CI from testing them. Raising the lower bound makes it formal, even though it's more for maintainers' convenience rather than a strong technical dependency.
I can revert this if you feel against it.

mtelvers: I think so, while the lower bound in opam-repository is 4.08, it's worth leaving this at 4.08.

PR #268, opened 2026-06-10
linux: add libzstd dev files correctly as done for debian.

frejsoya:
Unclear if the default should be to use libzstd. The current state is that Debian correctly adds the libzstd dev file, whereas Alpine and others do not.

6a6d0fa added the OCaml 5.1+ zstd `libzstd-dev` on apt but plain `zstd` everywhere else. That package is CLI-only on alpine/fedora/rhel/suse, so configure silently disables compression and switches come out ~90MB bigger than on debian (.cmt 91MB vs 33MB on alpine). Arch's `zstd` ships headers, left as-is.

Used opus-4.8

### DIFF
--- a/src-opam/linux.ml
+++ b/src-opam/linux.ml
@@ -57,7 +57,7 @@ module RPM = struct
 
   let ocaml_depexts v =
     if Ocaml_version.compare v Ocaml_version.Releases.v5_1_0 >= 0 then
-      Some "zstd"
+      Some "libzstd-devel"
     else None
 
   let add_user ?uid ?gid ?(sudo = false) username =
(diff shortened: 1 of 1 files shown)


mtelvers (approved):
Thank you for this contribution.

The diagnosis is spot-on: `zstd` is CLI-only on Alpine/RHEL/SUSE, so OCaml 5.1+'s configure silently skipped `libzstd` and switches came out ~90MB larger than Debian.

I pulled the containers and verified each changed package name installs from the default repos and actually delivers the header:

| Distro | Package | Installs | `zstd.h` | `libzstd.so` |
|---|---|---|---|---|
| Alpine latest | `zstd-dev` | ✓ | ✓ | ✓ |
| Fedora latest | `libzstd-devel` | ✓ | ✓ | ✓ |
| Oracle Linux 9 | `libzstd-devel` | ✓ | ✓ | ✓ |
| Oracle Linux 10 | `libzstd-devel` | ✓ | ✓ | ✓ |
| openSUSE Tumbleweed | `libzstd-devel` | ✓ | ✓ | ✓ |
| openSUSE Leap 15.6 | `libzstd-devel` | ✓ | ✓ | ✓ |

